PR #46, opened 2015-02-04
Fix missing external modules with .d.ts extension

csnover:
The TypeScript compiler will look for files with both `.ts` and `.d.ts` extensions when resolving external imports, so atom-typescript also needs to do this.

I didn’t check to see if the existsSync check is really necessary or if this could just be updated to double up the number of files in the map, but I did verify this fixes the problem where you `import foo = require('./foo')` and the file is `foo.d.ts`.


basarat:
perfect. Thanks!

csnover:
I think that was the quickest PR-to-commit I’ve ever experienced. Thanks! :)

basarat:
@csnover published as `v0.25.0` :+1: ;)

PR #150, opened 2015-03-05
Update to use TS1.5 services

csnover:
This PR updates to use one of the prerelease 1.5 compilers from Arnavion’s nightlies along with a few changes to services that are necessary for working with 1.5. Maybe this can just go on a separate branch until 1.5 is fully baked, but it seems like it’s far enough along now that we should start working with it.

I also note that 1.5 appears to be delivered along with a server for language services (tsserver) which could be more useful in the long term than having our own child language services process, but this PR just gets things fixed up to work again with 1.5.


csnover:
Note that there are quite a lot of changes in the JS files because the release versions haven’t been getting compiled using `grunt build`.

basarat:
:+1: Merge whenever you feel like we should release this. I leave that decision completely up to you :heart:

csnover:
I think it needs to wait till at least more of the ES6 module support is in place, since right now the support in master is bit of a mixed bag (imports with destructuring are supported but it crashes if you generate declarations and doesn’t support the export default syntax). I’ll keep updating this PR until everything seems stable.

basarat:
@csnover I'm going to join you on this branch. I'll start by bringing in master right now.

basarat:
closed in favor of : https://github.com/TypeStrong/atom-typescript/pull/179

Issue #123, opened 2015-02-24
Enable sourcemaps


basarat:
These will get checked in.
Update `tsconfig.json` + grunt.

basarat:
A simple flick of `tsconfig.json` doesn't cut it. Atom does some fancy js loading that breaks TS sourcemaps.

johnnyreilly:
I was just about to raise an issue saying that sourcemaps  don't seem to be being generated despite my settings:

```
{
    "compilerOptions": {
        "target": "es5",
        "module": null,
        "declaration": false,
        "noImplicitAny": true,
        "removeComments": false,
        "noLib": false,
        "sourceMap": true
    },
    "filesGlob": [
        "./**/*.ts",
        "../scripts/typings/**/*.d.ts",
        "!../scripts/typings/angularjs/angular-mocks.d.ts"
    ],
    "files": [
        // ...
    ]
}
```

Looks like you already have it in hand though!

basarat:
@johnnyreilly sourcemaps from settings works. Just pused an example: https://github.com/TypeStrong/atom-typescript-examples/commit/c442a5cbf760eb1806988f2eb3f379467e7b13ef

The settings you have posted is invalid in that `module:null` is invalid. If you don't want to specify the `module` just remove it:

```json
{
    "compilerOptions": {
        "target": "es5",
        "declaration": false,
        "noImplicitAny": true,
        "removeComments": false,
        "noLib": false,
        "sourceMap": true
    },
    "filesGlob": [
        "./**/*.ts",
        "../scripts/typings/**/*.d.ts",
        "!../scripts/typings/angularjs/angular-mocks.d.ts"
    ],
    "files": [
        // ...
    ]
}
```

This issue is for getting sourcemaps working inside atom so that we see *.ts files instead of *.js files inside *our atom-typescript* code. :heart:

basarat:
@johnnyreilly added an issue to track better error reporting for tsconfig : https://github.com/TypeStrong/atom-typescript/issues/138 Probably after I am done with the panels work

johnnyreilly:
Ah I see - without module: null we're getting sourcemaps as expected.  Thanks!  :+1: 

I don't know if it's related but the `format` doesn't seem to be having any effect. Doesn't matter what I put in - it's ignored.  Shall I raise that as a separate issue?

```
{
    "compilerOptions": {
        "target": "es5",
        "declaration": false,
        "noImplicitAny": true,
        "removeComments": false,
[...]

PR #156, opened 2015-03-06
Clean up build process to separate compiled content from raw content

park9140:
Removing node modules as it is bad practice to include this in a repository.  

Moving compiled content to 'dist' directory for better separation while developing. 

Including map files to make debugging simpler.

Update out of date libraries.


park9140:
@basarat, I think this makes it much easier to work on the code.  I have a few things I need to add before I can use this in the office.  After this I intend to attempt to rectify the 600+ ms startup time, and depricated api usage.

basarat:
Glad I didn't do a push. I'll merge and then commit my stuff :heart:

basarat:
> Including map files to make debugging simpler.

did it really help? I didn't see atom picking it up when I tried : https://github.com/TypeStrong/atom-typescript/issues/123

basarat:
Nope. Sourcemaps don't work : 
image

It's because our JS gets wrapped by atom in `(function (exports, require, module, __filename, __dirname) {` and `});`

So sourcemaps are no longer valid : 
image

I'd be more interested in adding TypeScript support directly into atom once I get a chance. Looking at the recent `6to5` support its not that hard to do.

basarat:
@park9140 also, we are building with atomts instead of `grunt` : https://github.com/TypeStrong/atom-typescript/commit/613435082239e8032acedcff964584527fb23ede 

`grunt` is now there just for backup (when things go bang during development).  Also note `const enum` handling is broken in the `1.4` language service : https://github.com/TypeStrong/atom-typescript/commit/613435082239e8032acedcff964584527fb23ede#diff-4e626b81616e50c89d68cba32c8d4c51L34 so if the last build isn't from `atomts` then the output would mismatch and there will be conflicts :+1:

csnover:
> @park9140 also, we are building with atomts instead of grunt : 6134350 

Nobody will know if the documentation is not kept up to date :) https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#building

basarat:
@csnover on it

basarat:
@park9140 if sourcemaps indeed don't work for you as well delete them from the repo, grunt file,tsconfig. Or give me the go ahead ❤️

park9140:
@basarat, will do,  I popped my debugger up and thought I was seeing them, but I'll take another look, maybe I had some weird setting, I played with adding direct typescript support before and may have enabled something that made them work.

Issue #45, opened 2015-02-03
Figure out if we have a `tsconfig.json` above us in the directory tree that does not account for the file we are opening


basarat:
And show a warning message to the user.

And create a separate program for this file.

lierdakil:
Doesn't seem relevant anymore.

PR #179, opened 2015-03-10
Typescript 1.5

basarat:
Things to do. An ever updating list: 

* [x] Wait for 1.5 
* [x] Watch the filesGlob | files for changes. We only watch "changes" not "addition / deletion" for now.


csnover:
Yes, there is no version of TypeScript available anywhere that has the necessary patches to make ES6 import/export work with declarations (https://github.com/Microsoft/TypeScript/pull/2139), so what’s committed to my branch is a custom build with that patch applied. https://github.com/Microsoft/TypeScript/pull/2197 was also necessary in that build but has since been merged, so 2139 is the only one I’m waiting for until getting a new official build. Watch https://github.com/Microsoft/TypeScript/issues/2242 for updates!

basarat:
@csnover alright. So I will not run `npm install` :heart`. I'll be doing new fixes in this branch just because I can't live without `let` anymore.

basarat:
@csnover I suspect the `reference` bugs you are getting a specific to `1.5`. Cannot do build `ctrl+shift+b` anymore. Fails with : 

image

basarat:
We are not doing fs checks for references. Only doing it for `imports`. I'll fix it tomorrow. Off for now.

basarat:
@csnover @park9140 I am feeling that having stuff that is not referenced from *files in tsconfig.json* is always going to result in tricky error cases : https://github.com/TypeStrong/atom-typescript/issues/180 and https://github.com/TypeStrong/atom-typescript/issues/45 

I don't see a graceful way of handling this. Any ideas?

basarat:
@csnover sorry #180 was fixable :heart:  

Still don't have an answer for what to do with #45 though.

basarat:
A bug for me : 
image this should not happen.

park9140:
@basarat,  I was planning on using https://www.npmjs.com/package/vinyl-fs in order to simplify this scenario.  You can populate a vinyl stream with the glob patterns and it will keep itself updated when you use the watch methodology.

basarat:
Having a quick look I think it's meaning of glob is different : order is significant. I am sure the gulp users are used to it but that is not what we promised with our filesGlob. Our is that it must match all patterns.

basarat:
I can definitely feel the speed up :heart:

basarat:
@csnover is it okay if I merge and publish this a new major version? 

There are lot of good performance fixes here that I'd like to get in the hands of *everyone*. Its been working like a charm for me.

csnover:
TS team was hoping to have 2139 landed by end of the week so can we maybe wait until Monday to see if that happens? Otherwise I don’t think I have any major objections.

basarat:
> Otherwise I don’t think I have any major objections.

Thanks I'll wait till monday :heart:

basarat:
For my reference really: 

I am still getting inconsistencies between what LanguageServiceHost thinks is the result and what the Atom Editor thinks is the result after an edit. Sample: autoformat this file  https://github.com/TypeStrong/atom-typescript-examples/blob/master/bigfile/bigfile.ts 

image

Here `143903` is the original text length, `144167` is the formatted text return from LS, then `144167` is the length that it sent as a edit by the buffer, `0, 143903, 144167` = `minchar,limchar,newtext.length` is what the Language Service Host gets, so No issues exist before this point. BUT LSHost then computes that the resultant text length is `144431`. This the exact same result from the old LSHost and the new LSHost so I blame `ts.computeLineStarts`. I think it is best if I just fall back to using `buffer` https://github.com/atom/text-buffer and see if issue just goes away.

basarat:
I've been using this over the weekend as well as on Monday. Its much better than what we have on master so publishing as a major version.

basarat:
I have merged into master. That's a good thing as I can ask people to just use master for the latest fixes. However I'm not going to publish till we have a valid `package.json`.

PR #283, opened 2015-04-20
Make keybinds more specific

park9140:
Keybind for quick fix was not specific enough to override the default core `editor:newline` rule
@basarat


basarat:
Awesome :rose:

basarat:
`v2.15.13`

PR #303, opened 2015-04-24
Improved preferred quote experience, and an option for module path suggestin.

vilic:
When that module path switch is on, paths not within the same directory of current working file will be shown as "~/some/path/to/project/root".


vilic:
atom-typescript-module-suggestion

lib/main/atom/atomConfig.ts
```diff
@@ -17,13 +17,19 @@ class Config {
             type: 'boolean',
             default: false
         },
+        modulePathToProjectRoot: {
+            title: 'Show module path suggestion relative to project root.',
+            type: 'boolean',
+            default: true
+        },
```
basarat: I will merge. But will separately make the default `false` post merge as I personally prefer *relative* to file (and not relative to root). 

Also people might be confused by project (here you mean `atom` project but vs developers might think `tsconfig` project).

lib/main/atom/atomConfig.ts
```diff
@@ -17,13 +17,19 @@ class Config {
...
         preferredQuoteCharacter: {
-            title: 'Preferred Quote Character',
+            title: 'Preferred quote character',
             type: 'string',
-            default: '"'
+            default: 'none'
```
basarat: I will make the *default* as `"`.

lib/main/atom/autoCompleteProvider.ts
```diff
@@ -109,6 +111,7 @@ loadSnippets();
 
 export var provider: autocompleteplus.Provider = {
     selector: '.source.ts',
+    inclusionPriority: 1,
```
basarat: Thanks for this. Did you run into any errors because of *not* having this?

lib/main/atom/atomConfig.ts
vilic: that's true, I agree it's better to have the suggestion relative to typescript project root.

basarat:
image

Doesn't seem to work on windows. 

Note : I have very mixed feelings about this. Reason is that I don't feel comfortable about autocomplete showing me *one* thing and then adding *another* thing. Please don't mind but I am going to remove this option. Would be great if you can log an issue before doing such awesome work as I understand it took your time and I hate to waste it. :rose:

basarat:
> . Would be great if you can log an issue

You don't have to log issues, sometime you just want to do the work as see how it goes. I was just mentioning that *I will* look at the issues and provide my opinions :heart:

vilic:
I am using windows. It won't change what's suggested if they (working file and the suggested module) are in the same directory.

It's okay we directly discussing them in merge request, I don't mind you not accepting them. And yes I just want do the work as see how it goes. :D

lib/main/atom/atomConfig.ts
basarat: Ah. I just got why you made it `none` (which basically means *preserve*).

lib/main/atom/autoCompleteProvider.ts
vilic: Ahh, I remembered that the fuzzy suggestions and accurate suggestions are mixed together (there used to be an issue about it? but I can't find it last night), and thought adding this fixed the order. But, it doesn't seem to be necessary.

basarat:
> It's okay we directly discussing them in pull request, I don't mind you not accepting them. And yes I just want do the work as see how it goes. :D

thank you :)

vilic:
:rose:

PR #312, opened 2015-04-29
Don't include the default lib when using the noLib flag

r2jitu:
The default lib shouldn't be included when the noLib flag is set. This is the same behavior as the tsc command: https://github.com/Microsoft/TypeScript/blob/master/src/compiler/program.ts#L155


basarat:
thanks :rose:

Issue #349, opened 2015-05-16
Pending requests are just growing and growing


markogresak:
First of all, I'm sorry about the title, but I couldn't find a better description of my problem.

The problem is when I have a TypeScript project open, the pending requests are just growing, none of them complete.

![growing pending requests](http://i.imgur.com/AgNjktQ.png)
*and so on for as long as you can imagine (currently it's at ~400)*

It first happened when I was trying out TypeScript + Angular2 where I first upgraded TypeScript to [`mhegazy/TypeScript`](https://github.com/mhegazy/TypeScript) fork and to `1.5.0-beta` right after it was released. 
I've temporarily stopped working on this project and came back today after installing all new updates, hoping they would help, but did not. 

I have no idea of what to look at, I'm posting my `tsconfig.json`, if there is anything else that will provide any help, please let me know. Note that `"files"` are missing tsd typings, I'm not sure if this is a bug or a feature.

`tsconfig.json`:

``` json
{
    "version": "1.5.0-beta",
    "compilerOptions": {
        "outDir": "dist",
        "target": "es5",
        "module": "commonjs",
        "declaration": false,
        "noImplicitAny": false,
        "removeComments": true,
        "noLib": false,
        "emitDecoratorMetadata": true
    },
    "filesGlob": [
        "./src/**/*.ts",
        "!./node_modules/**/*.ts",
        "!./dist/**/*.*",
        "!gulpfile.ts"
    ],
    "files": [
        "./src/index.ts",
        "./src/public/js/app.ts",
        "./src/server/static.ts"
    ]
}
```

basarat:
> Note that "files" are missing tsd typings, I'm not sure if this is a bug or a feature.

Not a feature. Might be a bug

basarat:
Can you share the project? Will help with repro.

markogresak:
This is the project that I'm currently working on: https://github.com/markogresak/color-wars

The other one is private, but I have added you as a collaborator.

basarat:
> Not a feature. Might be a bug

Looking at https://github.com/markogresak/color-wars the glob is not a bug. As you have `./src/**` and typings are *outside* i.e. `./typings` and don't match `./src/**`. 

I'm not getting the other issue either (nothing hanging). Will investigate more :rose: I'm rate limited at the moment so can't get tsd defs :D

markogresak:
Oh of course, I've completely forgotten about that.

[...]

PR #355, opened 2015-05-18
add a note on updating to readme

markogresak:
Related to my issuse in #349.


basarat:
Let me try get a repro first

basarat:
Better in the FAQ. Don't want to scare away new users. Although at this point they don't have an option of another atom package ... but they might just run away from atom altogether and don't want that.

Issue #112, opened 2015-02-20
Refactoring Wishlist


basarat:
- extract interface from class, module, or function
- copy type from variable
- implement stub (to enable "use first" programming flow).  doStuff(p1 : string) doesn't exist on MyClass - would you like to implement it?
- "Declare this" refactoring on an instantiated variable.
- Declare all refactoring on a constructor function  

Courtesy @nycdotnet https://github.com/TypeStrong/atom-typescript/issues/59#issuecomment-73228618

basarat:
Wishlist 

* [ ] rename file module (this is a big pain and the reason why we added `///ts:import` transform in the first place)

rogergl:
I think I would volunteer for 'extract interface from class, module, or function'

nycdotnet:
Feel free if you want a challenge.  That'd be great!

Diullei:
:+1:

rogergl:
BTW: Any ides how to handle settings for the quickfixes. e.g. sort methods names yes/no ?

nycdotnet:
There is already a settings dialog for atom-typescript, though I would strongly suggest trying to avoid anything that would require using it.

In the situation you've provided - extract interface - perhaps have the interface members come out in the order they're declared (if that can be determined) and provide a new refactoring "sort interface properties alphabetically".

rogergl:
The ordering was just an example. I think that it should be possible to customize the behavior of the quickfixes and that it should be a per project setting. I would prefer a section in the tsconfig.json file, but I do not know if this is possible and (or) allowed.

rogergl:
Just wondering how the file path in a Quickfix result is supposed to work. If I return the following result test2.ts is opened and then I get an unhandled promise exception with TypeError: Cannot read property 'transact' of undefined.

```javascript
 result = {
            span: {
                start: 0,
                length: 5
            },
            newText: "hello",
            filePath: "test2.ts"  // does not exist yet
};
```

basarat:
> Any ides how to handle settings for the quickfixes. e.g. sort methods names yes/no ?

Two solutions (neither implemented yet) : 

1.) Allow the quick fixes to add a UI (https://github.com/TypeStrong/atom-typescript/blob/674ea18a76d9765d35bf9359269b2e73d3029113/lib/main/atom/commands/commands.ts#L423) 

Basically run the quickfix code *in the ui thread* to generate a UI. Since creating atom UI's is a bit hard try not to do this :) 

[...]

PR #385, opened 2015-06-09
add quickfix for readonly properties

rogergl:
This is the first attempt to implement a quick fix to create read only properties


basarat:
Cool. Better if we make it a generic `CreateReadonlyProperty` quick fix (instead of a `CreatePropertyFromConstructorArg`). 

The code can remain as it is, supporting additional locations (besides constructor arg) can come later. Just change the name :)

basarat:
@rogergl awesome! thanks :rose: 

Github tip: People don't get notified about PR changes unless you leave a message (after your commit) :checkered_flag:

basarat:
Thanks! I can only imagine your satisfaction though. The first quickfix is truly empowering and fun to write :) 

wrapinreadonlyproperty

rogergl:
Yes it was a really nice experience. Any more ideas for some 'missing' quickfixes ?

nycdotnet:
https://github.com/TypeStrong/atom-typescript/issues/112

PR #388, opened 2015-06-11
feat(grammar): support brace at entity.name.function

armorik83:
This is necessary to me. thanks.


basarat:
sure why not :rose:

basarat:
`4.5.10`

PR #411, opened 2015-06-25
Error panel UI tweak

nycdotnet:
I updated the panel heading to disallow text selecting (make it more like a proper app).  You can still select text in the actual error area.  I also made the heading toggle on double-click (same behavior as single-click on the heading-fold button.  Please merge if you think this is cool.


basarat:
:candy: sweet. Released with `4.6.7` :rose: 

Tip: I personally toggle with `ctrl+;` followed by `enter`

Issue #397, opened 2015-06-19
Disable a popup message on successful compilation


helarqjsc:
It would be nice to have an option to disable this message, it seems a bit annoying to me. I mean, just the message that says "TS emit succeeded" should be possible to disable. It's nice to have a popup when you have a typo somewhere in your code, but when everything goes well, I would like the plugin to do its job silently. Showing a popup every time saving goes successfully isn't very useful. Can such an option be added to this plugin?

The message I'm talking about:
![Screenshot](http://i.imgur.com/pv2wdGb.png "Screenshot")

basarat:
> Can such an option be added to this plugin?

I would prefer a better *UI* for this. Basically I imagine a small icon on the screen that shows the status of the `js` (different, old, up to date)

helarqjsc:
How would it know if the js file is old or up to date after you have just opened the editor?

basarat:
> How would it know if the js file is old or up to date after you have just opened the editor?

good question. It can assume *not* to begin with ... or it can send an initial *compile* command right after initial load (its async and in the background anyways)  ... or something else

basarat:
closed with #415

PR #415, opened 2015-06-26
Add icon that represents the current file status and remove popup on successful emit

helarqjsc:
As was proposed by @basarat in #397
It removes the annoying "TS emit succeeded" message on every save and instead adds a small icon on the bottom panel that represents the current state of the compiled JS file. Initially it says "File might be outdated", because we don't know the state yet. If the file is successfully compiled and emitted, the status says "File is up to date". If the file has been edited, but not saved yet or a compilation error has occurred, it says "File is outdated". If there was a compilation error, but emit succeeded, it shows a popup just like before. 

Here what it currently looks like:
Screenshot

Few things that might need to be fixed:
- The status text takes quite a bit of space on the panel, it might be better to permanently show the icon and display text only on hover;
- "TypeScript: Build" command doesn't change the file status yet, only on save compilation does;
- It would be nice to have an option to compile files after the editor has been opened, so the file status is guaranteed to be up to date by default.

Thoughts?


lib/main/atom/onSaveHandler.ts
```diff
@@ -6,7 +6,8 @@ import atomConfig = require('./atomConfig'); ///ts:import:generated
 ///ts:import=parent
 import parent = require('../../worker/parent'); ///ts:import:generated
 
-import {errorView, show} from "./views/mainPanelView";
+import {getFileStatus} from "../atomts";
```
basarat: This will create a cyclic reference. I'll fix it post merge, but just wanted to point that out :rose:

basarat:
Great work. Thanks a ton!

basarat:
Thoughts:

> It would be nice to have an option to compile files after the editor has been opened, so the file status is guaranteed to be up to date by default.

We can do a *check* of the potential emit vs. a file at the emit location. This would be *really cool* https://github.com/TypeStrong/atom-typescript/issues/416 :rose:

lib/main/atom/onSaveHandler.ts
helarqjsc: It might be a silly question, but why didn't this cyclic dependency cause any problems while I was working on this piece of code? Shouldn't it have crashed Atom or something?

basarat: > Shouldn't it have crashed Atom or something?

Nope. Its because we only use items *off* of the export *from functions*. This means that when these functions *actually execute* the are defined. If we used these items at the root level they would have been `undefined`. 

More: https://nodejs.org/api/modules.html#modules_cycles

Issue #410, opened 2015-06-25
Better Syntax Highlighting


dgarzon:
Syntax highlighting seems to be broken, or there may be something wrong with Atom's One Dark Theme. Any feedback on this @basarat? If it is something on our end, I can help with the fix. Attached you will find an example of a .ts file I am working on: 

![screen shot 2015-06-25 at 5 02 16 pm](https://cloud.githubusercontent.com/assets/867990/8365762/69d074b0-1b5d-11e5-86f7-63d3ceb0b041.png)

basarat:
Seems to be missing styles in `one dark` (or us using wrong style names). Compare monokai with one dark:

# One Dark
![image](https://cloud.githubusercontent.com/assets/874898/8368331/32445be4-1bec-11e5-91b7-c95838e0f1c7.png)

# Monokai
![image](https://cloud.githubusercontent.com/assets/874898/8368334/3b349688-1bec-11e5-9283-26bead140265.png)

basarat:
Marked as up for grabs

Cu3PO42:
After experiencing the same bug and finding this issue, I dug around a bit. The keywords that lack highlighting all have the classes `.keyword.operator.ts`. E.g. Monokai sets the highlighting on `.keyword`, you can see the highlight just fine.

Other themes, such as One Dark, Atom Dark and Base16 Tomorrow Dark specifically set the color of `.keyword.operator` back to the normal text color. The classes `.keyword.operator` are also set on normal operators, such as `=`.

If I want these operators highlighted is probably a matter of preference and I get that different themes have different opinions about that. On the other side, I do believe keywords such as `interface`, `declare`, etc. are intended to be highlighted all the time (if you have syntax highlighting turned on).

That would mean atom-typescript is indeed using wrong classes for these keywords. I have never dealt with Atom grammars, but I've had my fair share of fun with Vim's grammars. I'll try and figure out which classes would be more appropriate and send a PR.

dgarzon:
Hey @Cu3PO42! I've been away all day, but after examining how both themes work I came to similar conclusions. I saw your fork and everything is starting to look great! I made one small addition so that `entity.name.module` is colored in a darker shade of orange following on One Dark's color scheme. Here is the code: 

```
.editor, atom-text-editor::shadow {
  .source.ts {
    .meta.function {
      .entity {
        &.name.module {
          color: hsl(29, 54%, 61%);
        }
      }
    }
  }
}
```

[...]

PR #418, opened 2015-06-27
[RDY] Improve grammar.

Cu3PO42:
Originally coming from TypeStrong/atom-typescript#410 I started improving the grammar. I stumbled upon more things than originally noted there, though and am trying to get as much fixed as possible.

I added TODOs for everything I noticed so far and that isn't handled completely. Unfortunately that's still quite a bit, but I thought I'd share my progress so far.

**EDIT**: All TODOs handled.


basarat:
That's a lot of work and code quality is :100:

There is also this : https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScript.YAML-tmLanguage and (I believe generated https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScript.tmLanguage) If you are curious more about these. 

Thanks a lot for this. I'd be happy to press the `merge` button whenever you want :rose:

Cu3PO42:
Thanks for the feedback @basarat! I'm happy to help. When I'm done here, I'll post some screenshots, squash the commits and mark this [RDY].

Cu3PO42:
@basarat Looks like I'm pretty done here. That was more work than I anticipated.

I'll also create a branch for the One Dark color scheme and add the few TS specific things here, reference that here and send a PR to them when this one is merged :)

nycdotnet:
Wow!

Cu3PO42:
Had a quick idea there... Guess I wasn't quite as done as I thought.

Cu3PO42:
And of course, I did forget a final thing...

basarat:
Let us know when it's good to merge and no more pushes are coming ;)

Cu3PO42:
@basarat I thought it was when I tagged it RDY, however after doing some development in TS, well, I found some cases I missed.

At this point it is most certainly bug free and can be merged. If I happen to find more things I missed, well, I'll open a new PR.

basarat:
:rose: Let me use it for one day :+1: before pushing

basarat:
function parameter (although correct), is a bit too aggressive : 

image

Perhaps its needs a nested recursion?

basarat:
Commented out offending rule for now so that it doesn't block remaining good stuff :rose: from release ;)

Cu3PO42:
The theme I used did actually not highlight function parameters, so I didn't see that. What would you consider to be the "correct" behavior? Match the braces, but not what's in between, match everything that doesn't recurse? And just for my own reference, that theme is Monokai, right?

basarat:
> And just for my own reference, that theme is Monokai, right?

yes

> . What would you consider to be the "correct" behavior? Match the braces, but not what's in between, match everything that doesn't recurse?

Actually `parameter` is different from `arguments`. Perhaps we can leave `arguments` unhighlighted (and let their natural tokenization take over) , which is what the js grammar does:

image


Note: The typescript ast also `arguments` vs. `parameters` which you can check using the ast visualizer on the following:

```
function foo(a){
}
foo('asdf');
```

Also note: I know less about how `cson` grammars work than you do at this point :)

Cu3PO42:
The distinction between arguments vs parameters in JS and other C like grammars was actually the first thing I was going to check when I got to my PC (early morning here). Seeing as it appears to be the way I expected, I'm all in favor of just deleting that specific rule.

Actually I only added it in the first place because the original TS grammar I started from would have "tried" to match that. But, yeah, just delete it :)

I would hope I know some things about TextMate grammars now! Though I had never seen one in my life before I started this.

basarat:
> just delete it

Done! Been using it all day and it looks much prettier. Published with a major version : `5.0.0` (the actual breaking change is that we don't create a tsconfig.json for you ... the user needs to trigger a command and we inform the user to do so).

Thanks for you hard work :rose: :100: :star:

Cu3PO42:
You're quite welcome! Glad it looks good and that my work can get out to everyone. I'm honestly not all that familiar with tsconfig files, but I don't quite get why this needs a special case... Why would you create a tsconfig file for CSON?

Not that it matters, I'm just curious.

basarat:
> Why would you create a tsconfig file for CSON

Sorry for the confusion, `tsconfig.json` has got nothing to do with cson.

I meant that, I didn't do a version major (from `v4` to `v5`) upgrade because of *this* PR. I did that because of some other breaking change : https://github.com/TypeStrong/atom-typescript/blob/master/docs/CHANGELOG.md#v500

Cu3PO42:
I see, I actually had that thought after looking at Changelog.md. Just a misunderstanding there, thanks for clearing it up!

basarat:
Have a minor request notice `public` on member property vs. function :

image

No hurry and feel free to say *no* :)

Issue #416, opened 2015-06-26
Instead of "no emit requested" would be great to show the actual JS vs. TS comparison


basarat:
refs https://github.com/TypeStrong/atom-typescript/pull/415

nycdotnet:
I do miss the ability to see the emitted JS in an easy manner.  While I don't care for the web essentials-style split screen, it is useful from time to time.  I wonder if it would be possible to enable some sort of keyboard shortcut to "pop" the emitted JS out in a floating window.  Perhaps a bit like a code lens preview.  This would be tricky because it would need to work with source maps.

nycdotnet:
@basarat Did you make a recent change to disable the "this file changed - do you want to reload?" prompt when looking at the emitted JS and TS in the same view as a split-screen?  Or was this a recent change in Atom?  Now it appears that the viewed JS file immediately refreshes when the TS compiles-on-save as long as you haven't made any changes in the JS file directly.

I notice we also have the same problem as Web Essentials used to where the emitted JS scrolls to the top when the file is refreshed.  If you didn't make this change then perhaps this is an Atom bug.

basarat:
> I do miss the ability to see the emitted JS in an easy manner

Already there : https://github.com/TypeStrong/atom-typescript#output-toggle

> where the emitted JS scrolls to the top when the file is refreshed. If you didn't make this change then perhaps this is an Atom bug.

Verified that this happens now. Its a atom bug nothing to do with us :rose:

nycdotnet:
> Already there : https://github.com/TypeStrong/atom-typescript#output-toggle

I do need to keep more up to date on the patch notes.

basarat:
fixed with https://github.com/TypeStrong/atom-typescript/pull/422 thanks to @helarqjsc :rose:

PR #422, opened 2015-06-30
Implement actual TS vs JS comparison

helarqjsc:
Refers to #416 

I think it would be better to add an option to disable this feature. Currently an emit and comparison are performed whenever a file is opened, which might be costly.


basarat:
> I think it would be better to add an option to disable this feature

Lets see how it goes before adding this feature. Should be *pretty fast* IMHO.

PR #431, opened 2015-07-02
[RDY] Further improve grammar.

Cu3PO42:
It was brought to my attention that a few things were added to TypeScript since I last used it on a larger scale and not all of these work properly right now. Specifically these are tuple types, indexable types and deconstructing in assignments. I'll try and get the things out there ASAP.

### DIFF
--- a/grammars/typescript.cson
+++ b/grammars/typescript.cson
@@ -35,7 +35,7 @@
     'patterns': [
       {
         'begin': '(?<=^|,|\\G)'
-        'end': '(,)|(?=;)'
+        'end': '(,)|(?=;)|(?<!,)$'
         'endCaptures':
           '1':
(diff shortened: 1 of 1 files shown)


basarat:
> I'll try and get the things out there ASAP.

No hurry. Do them at your *pleasure* :rose:

Cu3PO42:
@basarat I think we're good here, but you probably want to double check :)

basarat:
Merge at will and then `apm publish patch`

Also you might want to do `git config --global credential.helper wincred` if on windows to make workflow easier

Cu3PO42:
I'm pretty sure there are no bugs here, but I'll test for a bit longer and then merge!

And I'm on OS X, but thanks for the hint :)

PR #450, opened 2015-07-09
Fix vars and functions in grammar. Fixes #449.

### LINKED ISSUE #449: grammar issue
I noticed that while in a class if you use assign a variable without a semicolon and declare a method with a binding, say `my_method(): string { }` the method name and type "string" will not be highlighted

Example for you guys to plug in in case the issue was just for me. 

```
class Test {
    constructor() {
    }
    breaker() {
       let a = "I broke it"
    }
    inflicted(): string {
       // "string" doesn't have proper highlighting
    }
}
```

What it looks like for me before:
![before]
(http://i.imgur.com/nexhqCA.png)

And after:
![after]
(http://i.imgur.com/8NQh9Ez.png)

I fixed it by adding a "?" next to the semicolon in the typerscript.cson file on line 33. It seemed to fix it.
Line 33 before: `'end': '(;)|\\b(in)\\b|(?<!,|^)$`
Line 33 now: `'end': '(;?)|\\b(in)\\b|(?<!,|^)$'`

**I haven't done extensive testing to see if it covers everything, but it fixed the files that were not properly highlighted for me**

I would also like to add that it it seemed properly highlighted in v4.8.3, but I'm not entirely sure.

Cu3PO42:
Fix assignments with vars not terminating without a semicolon, new style functions not terminating in some cases and some old style ones not being highlighted correctly.


grammars/typescript.cson
```diff
@@ -295,6 +295,8 @@
   {
     'begin': '([\\p{L}\\p{Nl}$_.][\\p{L}\\p{Nl}$\\p{Mn}\\p{Mc}\\p{Nd}\\p{Pc}\\x{200C}\\x{200D}.]*)\\.([\\p{L}\\p{Nl}$_.][\\p{L}\\p{Nl}$\\p{Mn}\\p{Mc}\\p{Nd}\\p{Pc}\\x{200C}\\x{200D}.]*)\\s*(=)\\s*(?=(?:(?:async\\s+)?function)|(?:(?:<(?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/|\\|(?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/))+>)?\\((?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/|\\|(?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/))*\\)(?:\\:\\s*(?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/|\\|(?:[^&|*/+\\-\\)]|\\)(?=[\\[\\)|>])|\\/\\*\\*?(?:[^*]*?\\*)*\\/))+)?\\s*=>\\s*{))'
     'beginCaptures':
+      '0':
```
Cu3PO42: This change is necessary due to what I believe is actually a bug in Atom. This is functionally equivalent and shows correct behavior, though.

PR #457, opened 2015-07-10
Update file statuses on "TypeScript: Build" command

helarqjsc:
Currently a file status is only updated on save, after this PR build will also change it.
I've added `path.normalize` to `getFileStatus`, because on Windows it would sometimes get a path as "C:\xampp\htdocs\atom-typescript\lib\main\atomts.ts", sometimes as "C:/xampp/htdocs/atom-typescript/lib/main/atomts.ts" and they would be considered different files with different statuses. Now this should be fixed.
Also, during my test runs of "TypeScript: Build" some js files got updated. I haven't changed the corresponding ts files, so I guess it's fine and they have the right content now? It's weird that the js files weren't up to date in the master branch. Please, tell me if I shouldn't have pushed those files here.


basarat:
>  It's weird that the js files weren't up to date in the master branch.

The one is master is right. You need to run `npm install` after a pull to get the latest version of `ntypescript` that the project is using. What you are seeing is the danger of using a const enum (like the ts compiler does). 

I'll update contributing.

basarat:
I should have documented it as soon as I noticed it. I've done it now : https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#pull sorry for the annoyance :rose:

lib/main/atom/commands/commands.ts
```diff
@@ -99,6 +101,26 @@ export function registerCommands() {
...
+            let proj = getOrCreateProject(filePath);
+            proj.projectFile.project.files.forEach((path) => {
+                let status = getFileStatus(path);
+                status.emitDiffers = false;
+            });
+
+            resp.buildOutput.outputs.forEach((o) => {
+
+                // Emit never fails with an emit error, so it's probably never gonna be true
+                // It's here just in case something changes in TypeScript compiler
```
basarat: good call :rose:

lib/main/atom/fileStatusCache.ts
```diff
@@ -11,6 +13,7 @@ export interface FileStatus {
 
 let fileStatuses: { [index: string]: FileStatus } = {};
 export function getFileStatus(filePath: string): FileStatus {
+    filePath = path.normalize(filePath);
```
basarat: Prefer `consistentPath` https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/utils/fsUtil.ts#L5-L8 i.e. `/` *everywhere*. This is what the TypeScript compiler does *internally*. We have a bit of a dance as atom works one way, and typescript the other.

helarqjsc:
Ok, I'll fix the JS files in a moment and update this PR

dist/main/lang/fixmyts/quickFixes/addClassMethod.js
```diff
@@ -49,7 +49,7 @@ var AddClassMethod = (function () {
             typeString = getTypeStringForNode(binaryExpression.right, info.typeChecker);
         }
         else if (parentOfParent.kind == 165) {
-            var nativeTypes = ['string', 'number', 'boolean', 'object'];
+            var nativeTypes = ['string', 'number', 'boolean', 'object', 'null', 'undefined', 'RegExp'];
```
helarqjsc: Weird, why does it show this line as changed? Seems like it's exactly the same here in the master branch: https://github.com/TypeStrong/atom-typescript/blob/master/dist/main/lang/fixmyts/quickFixes/addClassMethod.js#L52

basarat: Weird indeed. Well ignore it. I'll merge locally to see what happens :+1:

lib/main/atom/commands/commands.ts
```diff
@@ -99,6 +101,26 @@ export function registerCommands() {
 
         parent.build({ filePath: filePath }).then((resp) => {
             buildView.setBuildOutput(resp.buildOutput);
+            let proj = getOrCreateProject(filePath);
```
basarat: Sorry I just re-read the code. *Don't call `getOrCreateProject` from the UI process*. This will double our memory consumption and make the UI much slower. Instead do this stuff in the `build` function's response (feel free to add a property there) and just use the response in the UI thread.

Some docs on architecture : https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#architecture 

Before I forget : Thanks for the PR! :rose:

helarqjsc:
Oh, I haven't realized `getOrCreateProject` is so memory intense. I removed it and use the response from `build` now, as you said. Sorry that you have to recheck this PR so many times :smile:

lib/main/lang/projectService.ts
```diff
@@ -86,6 +86,7 @@ export function quickInfo(query: QuickInfoQuery): Promise<QuickInfoResponse> {
 
 export interface BuildQuery extends FilePathQuery { }
 export interface BuildResponse {
+    project: project.Project;
```
basarat: Sending the `project` to the foreground is actually quite heavy (its a big tree, that needs to be JSON serialized). Better if you compute something like `tsFilesWithValidEmit` in the background and just send that to the UI process.

Thanks for updating the PR based on reviews :rose:

basarat:
> Sorry that you have to recheck this PR so many times

All good. Thanks for your continued time and effort :+1:

helarqjsc:
Now that I put more thought into it, it's probably better to return a list of all the files with invalid emit, rather than a list of files with valid emit, because TypeScript doesn't throw emit error anymore and every single file in the project is always valid after a build. Also, there is no need to update status of every file in the project, just the open ones, because when a file is open, its status is resolved by comparing existing js and a new emit, so a faster update of the opened files should be enough. I've updated the PR accordingly.

lib/main/lang/projectService.ts
```diff
@@ -86,6 +86,7 @@ export function quickInfo(query: QuickInfoQuery): Promise<QuickInfoResponse> {
 
 export interface BuildQuery extends FilePathQuery { }
 export interface BuildResponse {
+    tsFilesWithInvalidEmit: [string];
```
basarat: Think you mean `tsFilesWithInvalidEmit: string[];`

lib/main/lang/projectService.ts
```diff
@@ -162,7 +163,13 @@ declare module "${modulePath}"{
         fs.writeFileSync(defLocation, joinedDtsCode);
     }
 
+    let tsFilesWithInvalidEmit = outputs
+        .filter((o) => o.emitError) // only select outputs that produced an emit error
+        .map((o) => o.errors.map((e) => e.filePath)) // extract file paths from errors
+        .reduce((a, b) => a.concat(b), []); // flatten the array
+
```
basarat: Add `sourceFileName` to `EmitOutput` so that you don't need this map/reduce. Having duplicates in this list makes me uncomfortable ;)

helarqjsc: Sorry, I don't quite understand what you mean by adding `sourceFileName` to `EmitOutput`

basarat: I meant `filePath` here : https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/lang/modules/building.ts#L52-L57 

So that an `emitOutput` is associated with the `sourceFileName` easily and we don't need to do this lookup again later :rose:

helarqjsc:
Like so? :smile:

lib/main/atom/commands/commands.ts
```diff
@@ -99,6 +100,23 @@ export function registerCommands() {
 
         parent.build({ filePath: filePath }).then((resp) => {
             buildView.setBuildOutput(resp.buildOutput);
+            let editorsObserver = atom.workspace.observeTextEditors((editor) => {
```
basarat: You can use `atom.workspace.getTextEditors() ` instead of subscribing and unsubscribing :wink:

basarat:
Awesome :rose:

basarat:
`5.0.29`

PR #466, opened 2015-07-14
Adds support to generate Interface Declaration from JSON like data

xperiments:
This adds support to generate Interface Declaration from JSON like data.

Basically is a port from [json2dts](http://xperiments.in/json2dts/) and old project...

I have tried to "clone" the same code and logic found in the HTMLtoJSX, but I am not sure that populating the "main" directory with command dependencies is a good idea.


basarat:
> but I am not sure that populating the "main" directory with command dependencies is a good idea.

There aren't that many at the moment so its okay. Can be moved later :rose: 

Nice clean PR :wink:

basarat:
`v5.0.27`

Issue #469, opened 2015-07-16
Clarify over TypeScript Compiler version used / compileOnSave


johnnyreilly:
Hi Folks,

First of all, thanks for your hard work with atom-typescript - it's amazing.

A little background on where I am.  I'm just working out the best way to plug atom-typescript into my a current work project. This project is presently is written in React / ES6 and uses Babel with Browserify to transpile to ES5.  

So that everyone remains on board with the migration I need to tread delicately and respect certain approaches used in the codebase.  What this means is:
1. Do not use atom to do the TS compilation.  Instead this will probably be handled by @smrq's [TSIFY](https://github.com/smrq/tsify).  So atom-typeScript just provides the nice tooling but doesn't actually generate JS files.  I think [`"compileOnSave": false`](https://github.com/Microsoft/TypeScript/issues/2326) has my back here but I'm not certain - is that right?
2. Use a specify version of the TypeScript compiler (1.5-beta probably for now).  It's unclear to me which version of the TypeScript compiler atom-typescript uses.  Is it driven by the [`version`](https://github.com/TypeStrong/atom-typescript/blob/master/docs/tsconfig.md#version) in `tsconfig.json`?  Or is that just an aide memoire?

I've been reading up the docs but the answers to the points raised above aren't too obvious to me.  I'd be happy to update the atom-typescript documentation to clarify it once I know.

basarat:
> I think "compileOnSave": false has my back here but I'm not certain - is that right

Yes. (https://github.com/TypeStrong/atom-typescript/blob/master/docs/faq.md#i-dont-want-atom-typescript-compiling-my-js)

> Use a specify version of the TypeScript compiler (1.5-beta probably for now). It's unclear to me which version of the TypeScript compiler atom-typescript uses. Is it driven by the version in tsconfig.json? Or is that just an aide memoire?

No. Its ntypescript https://github.com/TypeStrong/atom-typescript/blob/master/package.json#L55  which is just a build of Microsoft/Master : https://github.com/TypeStrong/ntypescript

> I'd be happy to update the atom-typescript documentation to clarify it once I know.

Faq would be a good place (https://github.com/TypeStrong/atom-typescript/blob/master/docs/faq.md) to mention `ntypescript`. Although perhaps we will move to TypeScript nightlies which is something that the TypeScript team is planning :rose:

nycdotnet:
[...]

PR #475, opened 2015-07-18
Optional transpilation with Babel

nycdotnet:
I became despondent the other day when I saw that the `let` keyword's special scoping rules in `for`-loop iterators wouldn't be transpiled properly by TypeScript 1.5 to ES5.  https://github.com/Microsoft/TypeScript/issues/3915

While I am pleased that @danquirk seems amenable to implementing this in TypeScript, and I agree with @basarat that doing so [is the best long-term solution](https://github.com/TypeStrong/atom-typescript/issues/469#issuecomment-122177243), it could take some time.

Since this feature of `let` is transpiled properly by Babel today, I wondered if I could put together an enhancement for atom-typescript to optionally run the JS emitted by the TypeScript language service through Babel during the compile-on-save process.  As Bas predicted, it was actually fairly straight-forward, and this pull request is the result.

This PR utilizes a new optional key in tsconfig.json:
```json
{
    "externalTranspiler": "babel"
}
```

If key exists exactly as above, the emitted JS from the TS language service will be run through Babel; if not, it's not.

I would love your feedback on this.  Bas - would you consider merging?  I am open to any and all suggestions with regards to the config file, key names, etc.

**ToDo:**

- [x] Transpilation with Babel works based on the presence of a key in tsconfig.json.
- [x] Confirm source maps and other options are passed appropriately
- [x] Document
- [x] Test install/uninstall atom-typescript works as expected.  (Everything worked when I deleted the contents of my `%userprofile%\.atom\packages\atom-typescript\node_modules` folder and then ran `apm install`, so I am hopeful this is fine.


basarat:
@nycdotnet I'd be happy to merge and support this till typescript has emits for "let with closure" as well as "async await"

Till then typescript currently supports both these things with target es6. I think it provides great value to allow people to chain the typescript type checker and babel transpiler ie others have asked for it as well. So merge at will and thanks! 🌹

basarat:
And the name you have chosen seems to be worth supporting till infinity!

johnnyreilly:
+1 for this!

nycdotnet:
@basarat I believe that this is ready.  It actually seems to work shockingly smoothly for something I coded, especially with the fixes you made yesterday.  Would you please eyeball this and merge/release if you feel happy?

Note: I'm not quite sure of the best way to test uninstall/reinstall of atom-typescript without actually releasing it.  Are all the modules referenced in `package.json` always automatically downloaded by a package?

basarat:
> Are all the modules referenced in package.json always automatically downloaded by a package

node_modules? Yes. As a part of `apm update` which is what the users use (either manually or through the settings view). For us (`git clone` users), we need to run `npm install` as mentioned : https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#pull (could be improved on second look)

basarat:
:FANCY:

basarat:
Pushed with `5.2.0`. Lets see how it goes in userland :+1:

Issue #362, opened 2015-05-26
Transformers


basarat:
# Objective

Allow users to create custom syntax that is compiled down to TypeScript (and then js) and is type checked by the typescript compiler. 

Use Cases: React, Space-Pen, JSON, others?
# Design
## User Experience

User creates a `foo.tst` (typescript transformed) file. The file is handed off to the relevant transformer which spits out `.ts` content. 
- We do not feed this content to the file system but we do feed it to the language service as a _pseudo_ `.ts` file. 
## Tranformation sections

Sections relevant for transformation are determined by `/*something*/  ....content .... /*something*/` if `something` is registered as a transform. There are various proposals for `something` See contenders below.
## Errors

Error requests for the file are requested against a _psuedo_ .ts and reported against the pseudo ts file. If they click on the error we show them the psedo ts file. We prevent the user to press `ctrl+s` on the pseudo ts file. This pseudo ts file is live updated if they update the original src file. 
## Autocomplete

**Lower Priority** TypeChecking and readability is really what I am after here. Can be done using source maps and passing the current autocomplete trigger point to the mapped `ts` location and then to the TypeScript language service.
## Allow grammar integration

Transformers can provide a grammar implementation using a `tokenizeLine` API that works similar to how we currently implement the TypeScript grammar.
## Specifying Transformers

Eventually these can be driven by tsconfig.json. Currently these will be hard coded into atom-typescript.
## Contenders for transformation syntax
- `/*transform*/` paired with `/*transform*/` This is inspired by previous work on `jsx` [here](https://github.com/jbrantly/ts-jsx-loader).  < -ve people dislike comments.
-  keyword + backtick [here](https://github.com/Microsoft/TypeScript/issues/3022#issuecomment-99987650) < -ve backticks need escaping

```
jsx`<div>Hello ${ this.props.name }</div>`;
```
- `__name` [here](https://github.com/Microsoft/TypeScript/issues/3022#issuecomment-102636796) < -ve might conflict with actual variable names.
- My opinion:

```
transform:jsx{ 
/*code here */ 
}transform:jsx 
```
# Research
## Source maps
- Defacto lib : https://github.com/mozilla/source-map
- Good article : https://hacks.mozilla.org/2013/05/compiling-to-javascript-and-debugging-with-source-maps/
- list : https://github.com/ryanseddon/source-map/wiki/Source-maps%3A-languages,-tools-and-other-info
[...]

PR #483, opened 2015-07-22
[WIP] Enable out of context TypeScript file support

### LINKED ISSUE #368: Error finding files not in tsconfig.json
There is an error in the console when you attempt to view and edit typescript files in your project that are not listed in the files array in tsconfig.json:

image

image

If test2.ts is in my files array, then there is no errors, and everything works normally.

In my project this error prevents all intellisense and autocompletion in those files, as well as causes the AtomTS worker to take up 100% of my CPU (although it does not in the example I pasted).

I would like to exclude files like unit tests written in typescript from being compiled to javascript, because I do not need/want them to output JS files normally.

nycdotnet:
This is an attempt to resolve #368.  I need your help, @basarat or anyone else - even if it's to tell me I'm going in the wrong direction.  This works enough as a proof of concept, but it needs to be significantly improved I think to be pulled-in.

The measure of success will be:
  - [x] Can load out-of-context TypeScript files with no "error popups".
  - [x] atom-typescript will provide a single hard-coded error for any out-of-context TS file indicating that the developer should check their tsconfig.json.
  - [x] Basic grammar support should exist for out-of-context TS files (such as syntax colorization), but nothing that requires the language service (such as go-to-definition or autocomplete) has to work.
  - [x] Everything should work fast with no race conditions. (seemingly is so after 1c43b8b5fe0c4a17f6a2f5ebb78f56308669a14f)
  - [x] There should be no unusual errors in the console.

Concerns:
**I think these are all alleviated now**
  * Are you aware of a simple cache library in JS?  I want to essentially memoize if a file name is out of context for an arbitrary number of milliseconds - perhaps 3000.
  * I'm getting errors in the console, so I feel like I missed a few places where I need to check if the file is out-of-context.  I have been at this for a while and need a break, though.
  * There is not an easy way to ask the TypeScript language service if a particular file is in context or not.  Am I missing something?  Right now we have to ask it to do some work and then handle the exception.

Please let me know what you think.


basarat:
> Are you aware of a simple cache library in JS?

Just create a map? : https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/lang/utils.ts#L48-L56

This is a 5am feel like grabbing a glass of water message. More review when I actually wake up :rose:

nycdotnet:
Thanks.  I think I've found the "right way" to ask the language service if a file is out of context, so I may revert most of these changes.

nycdotnet:
I figured out how to ask the language service if a file is out of context and this allowed me to delete a lot of bad code.  I think this is in fair shape.  Errors in the console are gone.  I haven't tested it with any transformer files - those may still be broken.  Please let me know what you think!

basarat:
> I haven't tested it with any transformer files - those may still be broken

don't worry about them. That feature is on hold : https://github.com/TypeStrong/atom-typescript/issues/362 



> I think this is in fair shape

agreed. LGTM :+1:

nycdotnet:
I'm merging this now.  I will open a separate issue for this, but it seems that the user experience when creating a new .ts file in a blank folder is bad now.  There is an unhandled exception and the atom-typescript menu items are no longer available via Ctrl+Shift+P.

nycdotnet:
Well perhaps you fixed it - so nevermind.  It's working as expected now.  :-)

basarat:
> Well perhaps you fixed it - so nevermind.

yeah I missed up for second. Sorry about that

PR #491, opened 2015-07-25
Fix #481.

### LINKED ISSUE #481: String `") {"` breaks subsequent tokens
E.g. 

```ts
function write(str:string){
}

write(") {");
// var foo = 123;
```

Look like: 

image

Cu3PO42:
Add an extra rule to match function calls to override broken JS grammar.

screen shot 2015-07-25 at 15 19 36


Cu3PO42:
After a day of usage, I can't seem to find any further problems here.

Issue #507, opened 2015-07-30
Block comment: use // instead of /* */


markogresak:
I am setting up my atom typescript workflow again after formatting my computer and I can't figure out how to make `cmd + /` perform a block comment by adding `//` before each of line instead of creating an actual block comment.

What happens:
Initial state:
```
line 1
line 2
line 3
```
Comment block (`cmd + /`) action:
```
/*line 1
line 2
line 3*/
```
What I would like to do:
```
// line 1
// line 2
// line 3
```

I'm used to this from other grammars with this style of comments because if I want to uncomment only one line, I don't have to modify anything other line. If I would do the same for line 3, I have to change line 1 and line 3 first.

I am sure I had this functionality before, but I haven't been using atom-typescript for almost two months and now I can't find where to enable it again.

basarat:
> I am sure I had this functionality before, but I haven't been using atom-typescript for almost two months and now I can't find where to enable it again.

Probably broke post this : https://github.com/TypeStrong/atom-typescript/pull/418

Note the `.js` files behave that way in atom i.e. we get multiple `//`. 

@isuda has raised the same issue privately with me. Marking as a bug :rose:

markogresak:
Sorry for two referenced messages, I later amended commit message and github still recorded it as 2.

markogresak:
Since #508 was pulled, this is resolved.

Issue #159, opened 2015-03-06
Auto-indent doesn't work?


mindplay-dk:
The only problem I have with this plugin so far, is it seems to mess up auto-indent?

Using ATOM 0.186, with "Auto Indent" enabled, "Show Indent Guide" switched on, "Tab Length" set to 4 and "Soft Tabs" enabled.

Whenever I press ENTER in the editor, the cursor positions itself at the beginning of the line.

I have no other community packages installed, just this one.

PS: Thank you for this incredible plugin! I have tried every other Typescript plugin for ATOM (typescript-atomizer, language-typescript and atom-typescript-tools) some of which threw errors, none of which were able to even auto-complete. Excellent work!

Are you getting paid for this work, or where can we donate?

basarat:
> Are you getting paid for this work, or where can we donate?

No plan for a financial donation mechanism at the moment :heart:. It would be great if you can [leave a review on twitter](https://twitter.com/intent/tweet?text=Silky smooth TypeScript development&url=https://atom.io/packages/atom-typescript&hashtags=Atom,TypeScript&via=basarat)

ref : https://github.com/TypeStrong/atom-typescript/issues/66

basarat:
![autoindent](https://cloud.githubusercontent.com/assets/874898/6524712/c1c23b36-c450-11e4-9ac9-bc6bbaaf1a67.gif)

It doesn't work. We didn't do any *special* work to tweak to so it was expected. As you can see in the gif Atom is doing *something* for TypeScript files in that the brackets are placed correctly but you still need to indent the cursor into place. I suspect its done because of `scope` being `source` (we set the grammar scope to be `source.ts`). This magic done by atom is the reason why it hasn't been crucial. Going all the way with great auto-indent shouldn't be too hard if it goes according to my plan. Expect it soon.

basarat:
research : https://github.com/griiin/atom-auto-indent/blob/master/lib/auto-indent.coffee

basarat:
The non dynamic grammars get their indent settings from : https://github.com/atom/language-javascript/blob/master/settings/language-javascript.cson

basarat:
Based on my last commit `atom.config.get('editor.increaseIndentPattern', {scope: ['.source.ts']})` gives `"(?x) \{ [^}"']* $ | \[ [^\]"']* $ | \( [^)"']* $ | case[\s\S]*: $ | default: $"` 

[...]

PR #508, opened 2015-07-31
:bug: fix comment string as pointed out in #507

markogresak:
I have found out that the grammar relies on [`#comments`](https://github.com/TypeStrong/atom-typescript/blob/master/grammars/typescript.cson#L1292), which is in fact copied from [javascript grammar](https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson). 
Adding settings (copied from javascript grammar) ensures that the `commentStartString` is set to `// `.

The grammar options taken from javascript grammar, such as `#comments` I referred to earlier, might not even be necessary, since they are already provided from js grammar, but this has to be investigated and tested further, it should be done in separate PR.

### DIFF
--- a/settings/language-typescript.cson
+++ b/settings/language-typescript.cson
@@ -1,0 +1,12 @@
+'.source.ts':
+  'editor':
+    'commentStart': '// '
+    'foldEndPattern': '^\\s*\\}|^\\s*\\]|^\\s*\\)'
+    'increaseIndentPattern': '(?x)
+        \\{ [^}"\']* $
+      | \\[ [^\\]"\']* $
+      | \\( [^)"\']* $
+      '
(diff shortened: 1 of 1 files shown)


basarat:
That moment when the blame points back to yourself : 

I added this file after much investigation on a late night : https://github.com/TypeStrong/atom-typescript/blob/cca22b287e88de8057f7b81936efde1c5f132ffd/settings/typescript.cson (my comment https://github.com/TypeStrong/atom-typescript/issues/159#issuecomment-77554462) 

Then somehow decided that this file wasn't used : https://github.com/TypeStrong/atom-typescript/commit/424b4242040edceae93d13145ea962fb71b5237b 

:facepunch: myself. Sorry :rose: An thanks :rose:

Issue #542, opened 2015-08-25
Quick Fix overlay toggled in wrong file


Tharkun86:
Hi,

the Quick Fix overlay does only on the first file it is triggered. When triggered on a second file, the menu  will still appear in the first file it was triggered on and behave strangely (like staying open until a option is chosen or it is triggered again in the first file). 

I looked a bit into the code and will have a pull request ready for this shortly.

basarat:
Awesome, thanks! :rose:

basarat:
comes out with `5.5.1`

PR #543, opened 2015-08-25
Fix Quick Fix overlay issue when toggled on a second file

Tharkun86:
This fixes the issue mentioned in https://github.com/TypeStrong/atom-typescript/issues/542. When the SimpleOverlaySelectListView is called after the first initialization, only the options are set on the singleton but not the editor. This leads to the overlay opening in the previous editor when triggered from a second editor. 

This is simply fixed by setting the editor, too, when the singleton already exists. I tested this out for a while and seems to work fine.


basarat:
closes https://github.com/TypeStrong/atom-typescript/issues/542

Issue #524, opened 2015-08-10
For `compileOnSave: false` don't show the js emit status message


basarat:
As reported by @stijnvn :rose:

PR #546, opened 2015-08-29
Hide fileStatus when compileOnSave is false

wmaurer:
Fixes issue #524

Hope it looks okay. I'm pulling the compileOnSave directly from the cache.

### DIFF
--- a/lib/main/atom/views/mainPanelView.ts
+++ b/lib/main/atom/views/mainPanelView.ts
@@ -6,7 +6,6 @@ import atomUtils = require("../atomUtils");
 import parent = require("../../../worker/parent");
 import * as utils from "../../lang/utils";
 import { FileStatus, getFileStatus } from "../fileStatusCache";
-import { getOrCreateProject } from "../../lang/projectCache";
 
 var panelHeaders = {
     error: 'Errors In Open Files',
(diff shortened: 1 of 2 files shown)


lib/main/atom/views/mainPanelView.ts
```diff
@@ -205,14 +206,19 @@ export class MainPanelView extends view.View<any> {
 
     ///////////// Change JS File Status
     updateFileStatus(filePath: string) {
-        let status = getFileStatus(filePath);
-        this.fileStatus.removeClass('icon-x icon-check text-error text-success');
-        if (status.emitDiffers || status.modified) {
-            this.fileStatus.text('Js emit is outdated');
-            this.fileStatus.addClass('icon-x text-error');
+        let project = getOrCreateProject(filePath);
```
basarat: `getOrCreateProject` is something that shouldn't be called from the front end code. We have a seperate nodejs process running the background managing projects / language server etc. More docs : https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#architecture

wmaurer:
Ok thanks, that's clearer to me now. Is calling parent.getProjectFileDetails the correct approach?

basarat:
> Is calling parent.getProjectFileDetails the correct approach?

Yes :+1: However I would prefer you track down usages of `updateFileStatus`. I suspect they already have access to the `projectFileDetails`.  Look at https://github.com/TypeStrong/atom-typescript/blob/9a7680146e6b6ece5edaba3aebeeb9baa7b8a8cd/lib/main/atom/onSaveHandler.ts#L33 

Maybe add a function like `hideEmitStatus` there :rose:

wmaurer:
Unfortunately, not all other usages of `updateFileStatus` have immediate access to `projectFileDetails`, e.g.: https://github.com/TypeStrong/atom-typescript/blob/2a7b56303ae5b752fd35e0ae8d0950a895f520ee/lib/main/atomts.ts#L126
(in this case I could move the `updateFileStatus` up into the `then` on line 51)

And here I would have to fetch `projectFileDetails`, as it's not already called:
https://github.com/TypeStrong/atom-typescript/blob/2a7b56303ae5b752fd35e0ae8d0950a895f520ee/lib/main/atom/commands/commands.ts#L120

It seems to me to be a bit anti-DRY to have logic and code calling a `hideEmitStatus` wherever we currently call `updateFileStatus` (6 places at the moment), but I can do it if you like ...

Another idea would be to react to changes to the compileOnSave flag, and hide/show the emit status based on that. The updateFileStatus logic would still run for every file, but whether it's hidden or shown would be project-wide.

Let me know what you think.

basarat:
It will be in the next release 🌹

wmaurer:
Thanks!

PR #569, opened 2015-09-12
Conserves padding during autocomple in JSX

### LINKED ISSUE #525: padding removed when uses autocomple in jsx
Hello, @basarat!
I catch simple issue with autocomple in tsx. 
you can see pictures. 
start type div then space, then ctrl+space then press enter
screenshot_9
you get <divref instead <div ref
screenshot_2
Thank you!

gregerolsson:
By trimming the replacement prefix before handing it back to autocomplete-plus the whitespace in the user code is preserved when activating a suggestion.

Fixes #525

### DIFF
--- a/dist/main/atom/autoCompleteProvider.js
+++ b/dist/main/atom/autoCompleteProvider.js
@@ -51,7 +51,7 @@ exports.provider = {
                     var suggestionText = relativePath;
                     var suggestion = {
                         text: suggestionText,
-                        replacementPrefix: resp.endsInPunctuation ? '' : options.prefix,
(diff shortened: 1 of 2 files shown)


basarat:
Thanks :rose: Before I push is there any *known* edge case that I should know about. < always good to ask :wink:

gregerolsson:
I've tried breaking it in various ways, but it appears to be working correctly inside and outside of JSX. Any whitespace you have before selecting a suggestion is preserved.

autocomplete

basarat:
Awesome. Published with `6.0.6` :rose:

gregerolsson:
Thanks for all your hard work!

basarat:
> Thanks for all your hard work!

PR #585, opened 2015-09-18
Fixed bad format for some themes & status now easier to read

JohnMurga:
Status bar will corrupt/wrap with some themes and zoom levels, this fixes that
JS compilation text is more consistent with the other status text and easier to read as it is shorter


basarat:
sweet :rose: `6.0.10`

PR #607, opened 2015-09-23
Tidy up the CSS for the bottom panel

### LINKED ISSUE #606: Panel Sizing Breaks When Various Themes
image

I'll look into making some CSS adjustments to ensure it remains the same across theming.

blakeembrey:
Mostly making use of flex box instead of display hacks, but also made a change for the buttons to overlay the slightly larger panel resize drag area.

image

image

image

image

image

Edit: Moved the `showPending` to the parent element so people can click on the number (oddly, the theme I'm using doesn't even have a loading icon yet).

Closes #606

### DIFF
--- a/dist/main/atom/views/mainPanelView.js
+++ b/dist/main/atom/views/mainPanelView.js
@@ -29,23 +29,19 @@ var MainPanelView = (function (_super) {
         var btn = function (view, text, className) {
             if (className === void 0) { className = ''; }
             return _this.button({
-                'class': "btn " + className,
+                'class': "btn btn-sm " + className,
                 'click': view + "PanelSelectedClick",
-                'outlet': view + "PanelBtn",
-                'style': 'top:-2px!important'
+                'outlet': view + "PanelBtn"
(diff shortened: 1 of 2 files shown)


blakeembrey:
Actually, is the panel resize meant work?

basarat:
> Actually, is the panel resize meant work?

Nope. Its fixed height. Code was ported from https://github.com/tcarlsen/atom-message-panel and then gradually cleaned up (flexbox) but only as necessary (polymer layout styles weren't even in my sights back then)

LGTM :rose:

blakeembrey:
Great, I can merge this PR now. As for the panel resize icon, it's a little confusing that I can't drag (you'll know I tried). Should it be removed?

basarat:
> Should it be removed?

Yes please

blakeembrey:
Removed :+1:

basarat:
Merge then `apm publish patch`

`apm` works off your github credentials and it should just work for you :rose:

PR #635, opened 2015-10-02
Fix problem with syntax highlihtingting

prefiks:
So first a disclaimer, i almost completely don't know how those grammar files should be working, pretty much just tried some things until got results i wanted, it may broke some other things.

Those changes fixes two things that were bothering me, first missing syntax in atom 1.0.15+ and second change fixes missing missing syntax highlighting of jsx elements in some context (for example in `return <something/>` or `let v = <value/>`


basarat:
Phew ... Thanks :rose: This was driving me crazy, I guess you are breaking on newlines, and that fixes it?

/cc @Cu3PO42

basarat:
release with `7.1.0`

subjectix:
@prefiks Thank you.

prefiks:
I think problem with changed rules was that both begin and end regexes could match zero width string, so they could not advance current position. Optimization added in 1.0.15 didn't like that, and caused problem with highlighting.

basarat:
Note to self : http://www.apeth.com/nonblog/stories/textmatebundle.html

grammars/typescript.cson
```diff
@@ -711,12 +711,12 @@
...
     'patterns': [
       {
-        'include': '$self'
+        'include': '$base'
       }
     ]
 
   'type':
-    'begin': '(?=)'
+    'begin': '(?=[\\(\\{\\p{L}\\p{Nl}$_])'
```
basarat: I'm unable to understand how this code is designed to work. 

See https://regex101.com/r/qO2zF7/1

image

What's up with matching `p` or `{L}` etc. ... or am I misreading this. Thanks for the fix btw. Perhaps all we needed was literals `(` and `{`?

blakeembrey: In some regexp implementations, that would work for internationalisation categories - matches letters or number letters.

Reference: http://xregexp.com/plugins/, no idea if Atom supports it personally, but I know they do fix some back references and other regexp features.

basarat: Makes sense now https://regex101.com/r/cZ1zZ6/1 . Cheers :rose:

Issue #377, opened 2015-06-03
Support the `exclude` property


basarat:
* Update to latest master
* Support the property 
* Update the default generated tsconfig to be "exclude" only

https://github.com/Microsoft/TypeScript/pull/3188

Need to think about its interaction with `filesGlob`

elmariofredo:
:+1:

basarat:
People have expressed desire to still have `filesGlob` so I'll wait on this for a while

nycdotnet:
It seemed that the only reason the TypeScript team went with exclude is that they didn't want to take a dependency on a glob matcher library.  It would be a shame to lose this great functionality.  I hope filesGlob doesn't have to go away when exclude support is implemented.

Cu3PO42:
I strongly agree with everyone who likes `filesGlob`. Even though `excludes` gives you some of the power, you may want to exclude all TS files in a certain folder and not list them manually. This might be the case if you have copies of a dependency which is subject to change.

basarat:
Updated issue accordingly :rose: `filesGlob` will stay

Cu3PO42:
I'm very glad to hear that :+1:

basarat:
Up for grabs

# Implementor guidance 

Source code needs to go here : [tsconfig.ts](https://github.com/TypeStrong/atom-typescript/blob/1f516fd2061fe2e6dcaea4fbc560024fdd35121d/lib/main/tsconfig/tsconfig.ts)

For guidance one can use the work done [here](https://github.com/TypeStrong/tsconfig/blob/be5007c5505f73810c6510ecbdc166bed4dfe036/src/tsconfig.ts). Basically add the excludes to the pseudo filesGlob.

tanguylebarzic:
@basarat  Please have a look at https://github.com/TypeStrong/atom-typescript/pull/637. I'm happy to work on improving this pull request.

basarat:
closed with #637

Issue #332, opened 2015-05-07
File globbing slow with large node_modules folder


poelstra:
Atom Typescript adds the following `filesGlob` to `tsconfig.json` by default:

```
"./**/*.ts",
"!./node_modules/**/*.ts"
```

This indeed excludes things from `node_modules`, but it apparently still causes the full `node_modules` folder to be traversed.

I've tried replacing the second line with `"./node_modules/**"` and `"./node_modules/"` in order to prevent it from entering the directory at all, but that didn't seem to help (or didn't work at all).

We have a very large number of node modules in this particular tree, which makes Atom very slow for many operations (adding a file, renaming) and causes frequent hickups while editting. Haven't timed it exactly, but it's in the order of tens of seconds before e.g. a save of tsconfig.json 'notices' added or removed files.

As an experiment, I moved the `tsconfig.json` to the subdir that actually contains the TS code (which doesn't have a node_modules dir), and now saving tsconfig,json is 'just' a second or two (it's a fairly large codebase). 

Somewhat related to #217, which also got slow due to traversing large number of directories.

Is there e.g. a different glob pattern that may help?

poelstra:
Seems Microsoft is also trying to do something similar: https://github.com/Microsoft/TypeScript/issues/3043

basarat:
can you try putting the exclude first: 
```
"!./node_modules/**/*.ts",
"./**/*.ts"
```

poelstra:
Tried these (note: we need the .d.ts line too, as we generate the .js etc next to the .ts in this project):

```
        "!./node_modules/**/*.ts",
        "./**/*.ts",
        "!./**/*.d.ts"
```

```
        "!./node_modules/**",
        "./**/*.ts",
        "!./**/*.d.ts"
```
```
        "!./node_modules/**/*",
        "./**/*.ts",
        "!./**/*.d.ts"
```

They are all still slow, and worse, they add .ts below node_modules files to the list.

Also tried
```
        "!./node_modules/**/*.ts",
        "./**/*.ts"
```
just to see if it was our last .d.ts line, but that still leads to stuff in node_modules to be included (and it's also slow).

nycdotnet:
Out of curiosity, could you just do '!./node_modules' ?

poelstra:
I'm not at my work PC right now, but I think I tried that too and that it didn't work.

It was a bit cumbersome to test though, given that I typically had to wait a long time and often needed to kill Atom, so I may have overlooked something.

[...]

PR #637, opened 2015-10-02
Add support for exclude in tsconfig.

tanguylebarzic:
Aimed at closing https://github.com/TypeStrong/atom-typescript/issues/377. However, please note that contrary to https://github.com/Microsoft/TypeScript/pull/3188, this support globs. Also, this seems to be too simple... It does solve our usecase (excluding node_modules, same issue as https://github.com/TypeStrong/atom-typescript/issues/332).


basarat:
> please note that contrary to Microsoft/TypeScript#3188, this support globs

Lets not document that it supports it :) Its an implementation detail...as long as it supports *what is meant to be supported* it is okay :rose:

PR #649, opened 2015-10-08
Move to using `glob` library (optimised globbing)

blakeembrey:
@basarat You might want to try this out quickly, but it's much better in my test case.

Edit: As in, I have no idea how to test/log stuff related to this to make sure it's correct.


basarat:
Awesome ! :rose: 

Sorry for including `node_modules` originally ... I was pushing hard and from various places so was a bit lazy back then. Never again

blakeembrey:
No worries, there should be a bit more checks on this, but you're more experimental than me now :smile: If this works for you, I can merge the same PR in `tsconfig` :stuck_out_tongue_closed_eyes:

basarat:
https://github.com/TypeStrong/atom-typescript/releases/tag/v7.3.0 :rose: 

Again thanks a lot for this!

PR #653, opened 2015-10-09
chore(rebuild) rebuild with self :rose:

basarat:
@blakeembrey do you get this diff if you rebuild on master as well? 

It works fine (tried vs code) so no biggie :rose: maybe its a windows vs. mac issue? I'll investigate more if you don't get the same diff.


blakeembrey:
@basarat Sorry, that's definitely my fault on the `tsconfig` change. Since it was always (maybe we can change the API later?) outputting absolute paths, I have code there which is making it relative again and normalizing the path slashes. So yeah, it's going to result in that change.

basarat:
> So yeah, it's going to result in that change.

no worries :rose:

Issue #666, opened 2015-10-16
jsx highlighting breaks if text in span contains key words of typescript


stepancar:
Hello!
this screen demonstrate this issue
![screenshot_7](https://cloud.githubusercontent.com/assets/2854910/10541215/ecfff942-7417-11e5-914a-28900d6b7999.png)
Thank you!

basarat:
Sample :rose: 

```jsx
let x = <div>
    I do declare that this is a bug
</div>
```

![image](https://cloud.githubusercontent.com/assets/874898/10549613/e2de7954-748d-11e5-961f-30463db6eb56.png)

nycdotnet:
Good show old chap!

stepancar:
@basarat, maybe you can help me with this issue. I dont know how to fix it, and how to find reason of this behavior. But i wanna help. Thank you.

basarat:
> dont know how to fix it, and how to find reason of this behavior. But i wanna help. Thank you.

# Basic 

The issue is that the `tsx` grammar is breaking into `ts` in a string literal. 

# Further info

This is the grammar file : https://github.com/TypeStrong/atom-typescript/blob/master/grammars/tsx.cson which is a shameless copy of atom-react ([current code](https://github.com/orktes/atom-react/blob/249e0fab8e540d55ed84a13b41c7aa9518146655/grammars/JavaScript%20(JSX).cson)). If the bug exists here there is probably a bug upstream and worth reporting to https://github.com/orktes/atom-react/issues (or check if its already there) so that we get more :eyes: on it. 

Also  @Cu3PO42 is the resident grammar expert so one can direct questions for him :rose:

Also I have some docs linked here : https://github.com/TypeStrong/atom-typescript/blob/master/docs/grammar.md :rose:

PR #672, opened 2015-10-20
updated grammar for tsx

stepancar:
Hello @basarat! Yesterday i fixed [#666](https://github.com/TypeStrong/atom-typescript/issues/666), but it breaks
this grammar supports typescript (shameless copy [from  sublime-typescript](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScriptReact.YAML-tmLanguage) it's better then babel grammar because for typescript :)
Please, check

### DIFF
--- a/grammars/tsx.cson
+++ b/grammars/tsx.cson
@@ -1,137 +1,941 @@
 name: "TypeScript (TSX)"
-scopeName: "source.ts.tsx"
+scopeName: "source.tsx"
 fileTypes: [
 	"tsx"
 ]
 patterns: [
 	{
-		include: "#pcdata"
-	}
-	{
-		include: "#literal-jsx"
-	}
-	{
-		include: "source.ts"
+		include: "#expression"
 	}
 ]
 repository:
-	pcdata:
-		patterns: [
-			{
-				include: "#pcdata-quotes"
-			}
-		]
-	"pcdata-quotes":
-		name: "meta.other.pcdata.js"
-		match: "\\b\\w+[\"]\\w+\\b"
-	"literal-jsx":
-		contentName: "meta.jsx.js"
-		begin: "(?<=\\(|\\{|\\[|,|&&|\\|\\||\\?|:|=|=>|\\Wreturn|^return|^)(?=\\s*<[_$a-zA-Z])"
-		end: "(?<=/>|>)"
-		patterns: [
-			{
-				include: "#jsx-tag-start"
-			}
-		]
-	"jsx-tag-start":
-		patterns: [
-			{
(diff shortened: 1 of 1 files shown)


basarat:
@stepancar awesome ... I've been meaning to be lazy and just let MS handle the grammar... how are you doing the conversion ... can you document the steps so I can keep it updated? :rose:

stepancar:
document the steps? I just downloaded [sublime-typescript grammar](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScriptReact.YAML-tmLanguage) , then converted to json [usung this tool](https://www.npmjs.com/package/yaml-to-json) then converted to cson using [thist tool](https://github.com/bevry/cson) then just puted to atom-typescript, then smiled))

basarat:
Released with https://github.com/TypeStrong/atom-typescript/releases/tag/v7.4.4 :rose:

basarat:
Colors (and tokens) are bit different between `tsx` and `ts` : 
image

:-/ Guess we might need to change a few token scope names

blakeembrey:
If you generated one from Sublime, why not just do the other too? Or has there been a lot of custom improvements over Sublime? I'm wondering, also, if someone could bring this into a common core for grammars in TypeScript to avoid this kind of duplicate work.

basarat:
> I'm wondering, also, if someone could bring this into a common core for grammars in TypeScript to avoid this kind of duplicate work.

Agreed ... we have very little custom work ... and I moved that up the top : https://github.com/TypeStrong/atom-typescript/blob/729b5522cc17fc84511c05ebb5ae1812f3ec4b00/grammars/typescript.cson#L12-L80 

But I don't want to move `.ts` grammar over till everyone is happy with the token names in `.tsx` grammar (which we just pushed/released)

stepancar:
@basarat,
as i understand you should commit [your ts grammar changes](
https://github.com/TypeStrong/atom-typescript/blob/729b5522cc17fc84511c05ebb5ae1812f3ec4b00/grammars/typescript.cson#L12-L80)
to tsx grammar too (for tsx highlighting)

stepancar:
> I've been meaning to be lazy and just let MS handle the grammar

 &
> I'm wondering, also, if someone could bring this into a common core for grammars in TypeScript to avoid this kind of duplicate work.

Did MS have plans to handle grammar?

blakeembrey:
> Did MS have plans to handle grammar?

Not that I know of, but we can always ask. If converting is as simple as a basic transform, then we can probably look at automating the custom patches and contributing syntax highlighting patches upstream instead - otherwise `atom-typescript` will continue diverging anyway.

Issue #673, opened 2015-10-21
intellisense for tsx files not working


ra0o0f:
hi

after updating to 7.4.4 intellisense not working with .tsx files, snippet not working too.
ts files are ok

i tried to uninstall and reinstalling plugin, but it doesn't help either.

thanks

samuel-vincent:
Having exactly the same issue.

ra0o0f:
for now stay with 7.4.3

apm install atom-typescript@7.4.3

blakeembrey:
@ra0o0f @samuel-vincent I just pushed a quick patch for the scope of completion and linting. Can you let me know if it works now?

ra0o0f:
@blakeembrey code completion works now, but snippet doesnt yet.

exapmle: `importr` or `import`

![image](https://cloud.githubusercontent.com/assets/2182530/10646946/c6505c90-7842-11e5-968e-67fc651c60ba.png)

blakeembrey:
Unless I'm mistaken, snippets have never been supported there. I'll open a new issue for this.

blakeembrey:
@ra0o0f Support for snippets added with `v7.4.6`

ra0o0f:
@blakeembrey thanks a lot

PR #674, opened 2015-10-21
fixed auto completion

stepancar:
fix for [yesterday changes](https://github.com/TypeStrong/atom-typescript/pull/672) 
and for [this issue](https://github.com/TypeStrong/atom-typescript/issues/673)


blakeembrey:
I just pushed a patch to update the scopes, rather than change the grammar. Let me know if it works for you.

PR #677, opened 2015-10-21
workerLib's spawn should preserve the environment

Ptival:
Hello,

Working on NixOS, it is extremely bad that atom-typescript overwrites the environment of the process when calling spawn.
In particular, the child process fails to spawn as it cannot find libgtk-x11.

This patch attempts to solve this issue by only overriding the existing environment.

What is the policy with regards to pull requests? I notice the dist/ repository is tracked, which is weird, but I guess it's useful for bootstrapping?

Best regards,

### DIFF
--- a/lib/worker/lib/workerLib.ts
+++ b/lib/worker/lib/workerLib.ts
@@ -231,10 +231,16 @@ export class Parent extends RequesterResponder {
     /** start worker */
     startWorker(childJsPath: string, terminalError: (e: Error) => any, customArguments: string[]) {
         try {
+            var spawnEnv = process.env;
+            spawnEnv['ATOM_SHELL_INTERNAL_RUN_AS_NODE'] = '1';
             this.child = spawn(this.node, [
             // '--debug', // Uncomment if you want to debug the child process
                 childJsPath
-            ].concat(customArguments), { cwd: path.dirname(childJsPath), env: { ATOM_SHELL_INTERNAL_RUN_AS_NODE: '1' }, stdio: ['ipc'] });
+            ].concat(customArguments), {
+              cwd: path.dirname(childJsPath),
+              env: spawnEnv,
+              stdio: ['ipc']
+            });
 
             this.child.on('error', (err) => {
                 if (err.code === "ENOENT" && err.path === this.node) {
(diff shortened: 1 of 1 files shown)


lib/worker/lib/workerLib.ts
```diff
@@ -231,10 +231,16 @@ export class Parent extends RequesterResponder {
     /** start worker */
     startWorker(childJsPath: string, terminalError: (e: Error) => any, customArguments: string[]) {
         try {
+            var spawnEnv = process.env;
+            spawnEnv['ATOM_SHELL_INTERNAL_RUN_AS_NODE'] = '1';
```
basarat: Issues: 
 
a.) You are mutating `process.env` here ... that is bad
b.) Using `process.env` as a whole breaks using *atom as node* on windows. See : https://github.com/atom/atom/issues/2887#issuecomment-93887882 and independently verified : https://github.com/atom/atom/issues/2887#issuecomment-93893894

Solutions :

* Use `Object.create` around process.env
* only do if platform is `NixOS`

Thanks for finding the bug :rose:

basarat:
> What is the policy with regards to pull requests? I notice the dist/ repository is tracked, which is weird, but I guess it's useful for bootstrapping?

Its because atom packages are git tags ... and if the `.js` resources aren't there in github package will not work :-/ :rose:

lib/worker/lib/workerLib.ts
Ptival: Oh yes, oops! I will fix that.

Do you have an existing way of checking the operating system?

basarat: > Do you have an existing way of checking the operating system?

`process.platform` https://nodejs.org/api/process.html#process_process_platform although I don't know the value for NixOS .... I'd be okay with whatever that is as long as its not mac / windows :)

Ptival: It has value 'linux' under NixOS.
I can set it that way, though I am not able to test on many such operating systems.

basarat:
Released as a minor update : https://github.com/TypeStrong/atom-typescript/releases/tag/v7.5.0 :rose:

PR #683, opened 2015-10-24
Added notes on exclude property.

kenhowardpdx:
@blakeembrey @basarat A change that I think is important so devs understand `exclude` isn't a fix-all.

### DIFF
--- a/docs/tsconfig.md
+++ b/docs/tsconfig.md
@@ -27,7 +27,7 @@ i.e. an empty JSON file at the *root* of your project :heart: This will be suffi
 ### exclude
 These references are relative to the `tsconfig.json` path. This does not accept glob formatting.
 
-NOTE: `files` can override the `exclude` array. It's recommended to use either `exclude` & `files` -- OR -- `filesGlob`, not both. `filesGlob` solves the exclude problem as long as your team is exclusively using atom-typescript.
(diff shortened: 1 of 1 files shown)


docs/tsconfig.md
```diff
@@ -23,6 +24,25 @@ i.e. an empty JSON file at the *root* of your project :heart: This will be suffi
 
 ## Examples
 
+### exclude
+These references are relative to the `tsconfig.json` path. This does not accept glob formatting.
+
+NOTE: `files` can override the `exclude` array. It's recommended to use either `exclude` & `files` -- OR -- `filesGlob`, not both. `filesGlob` solves the exclude problem as long as your team is exclusively using atom-typescript.
```
blakeembrey: Using `files` **does** override the `exclude` array, you can not use them together. Probably help to be specific (https://github.com/Microsoft/TypeScript/wiki/tsconfig.json#details). For the logic used to support `filesGlob`, it's here: https://github.com/TypeStrong/tsconfig/blob/master/src/tsconfig.ts#L167-L187. It basically looks like:

1. Use `files` when `files` exists and `filesGlob` does not exist
2. Use `filesGlob`
3. When `exclude` is set, append exclusions to `filesGlob`
4. When only exclusions exist, match all TypeScript files

kenhowardpdx: I totally space the mutual exclusivity of `files` and `exclude`. Thanks for the gentle reminder. I'll make change.

kenhowardpdx: @blakeembrey I haven't been able to see # 3 working.

blakeembrey:
LGTM :+1:

PR #715, opened 2015-11-11
Support rewriteTsconfig flag in tsconfig.json

### LINKED ISSUE #703: Option to prevent rewriting tsconfig.json file
For people that don't care about keeping `files` in sync (for other IDE use) or just don't want us messing with their `tsconfig.json` file formatting, provide an option to disable tsconfig.json overwriting

* [x] Add some place to use this setting (the simplest location would be to use `tsconfig.json` as that file is available to `tsconfig.ts` (our code that rewrites `tsconfig.json`)
* [x] Add code to `tsconfig.ts` to use this option
* [x] Document option  [here](https://github.com/TypeStrong/atom-typescript/blob/master/docs/tsconfig.md) and optionally [here](https://github.com/TypeStrong/atom-typescript/blob/master/docs/faq.md)

sszuflita:
The rewriteTsconfig flag enables users to prevent Atom-Typescript
from rewriting their tsconfig.json. The flag defaults to true,
so users who upgrade won't observe any different behavior.

The FAQ and tsconfig docs have been modified to reflect
rewriteTsconfig.

This resolves #703.

### DIFF
--- a/dist/main/tsconfig/tsconfig.js
+++ b/dist/main/tsconfig/tsconfig.js
@@ -174,7 +174,8 @@ function getDefaultInMemoryProject(srcFile) {
         formatCodeOptions: formatting.defaultFormatCodeOptions(),
         compileOnSave: true,
         buildOnSave: false,
-        scripts: {}
+        scripts: {},
+        atom: { rewriteTsconfig: true },
     };
     return {
         projectFileDirectory: dir,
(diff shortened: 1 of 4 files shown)


basarat:
@sszuflita Thanks! Can you move it under the `atom` key. I know this is the first time we move stuff in there, but this is an option likely to be used by other things like grunt too.

sszuflita:
What do you mean by this? I am interpreting this to mean:

{
  "compilerOptions": {...},
  "filesGlob": [...],
  "files": [...],
  "exclude":[...],
  **"atom": {
    rewriteTsconfig: false;
  }**
}

If we store this option in an object labelled atom, how does this help grunt access it? To me it seems like storing it in an object labelled atom would would suggest this option is relevant only to atom.

basarat:
Yes I want it under atom. So that grunt can still rewrite if users want grunt to rewrite. Grunt would look at "grunt//rewrite..."  🌹

basarat:
Awesome. Thanks a million @sszuflita ! 

TIP: github doesn't notify if you only push more commits. Please leave a message in PRs that have been reworked :rose:

basarat:
Released https://github.com/TypeStrong/atom-typescript/releases/tag/v7.10.2 :rose:

PR #727, opened 2015-11-16
Added support for running the local copy of babel for a given project

david-driscoll:
This removes the dependency on `atom-typescript`'s local copy of babel, and allows a given project to use whatever version they like (assuming the core  `.transform` api doesn't change anyway).

Also adds support for loading configuration from `.babelrc`

### DIFF
--- a/dist/main/lang/modules/building.js
+++ b/dist/main/lang/modules/building.js
@@ -3,7 +3,9 @@ var path = require('path');
 var fs = require('fs');
 var fsUtil_1 = require("../../utils/fsUtil");
 var utils_1 = require("../utils");
-var babel;
+var findup = require('findup');
+var babels = {};
+var babelConfigs = {};
 exports.Not_In_Context = "/* NotInContext */";
 function diagnosticToTSError(diagnostic) {
     var filePath = diagnostic.file.fileName;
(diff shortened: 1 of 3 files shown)


basarat:
@david-driscoll big fan of yours. Just added you as an owner. Do and support whatever feature you want in this project all the way up to releasing new versions :rose:

 Let me know if you have any questions.

david-driscoll:
@basarat big fan of yours too! :100: 

I'm actually running into an strange issue here... when running with `debugSync` it works correctly, then when running as a process it fails... silently... right at `babel.transform`... now to figure out what's happening.

basarat:
> when running as a process it fails... silently.

I've had stuff like that happen. Running atom as node can be weird: https://github.com/Microsoft/TypeScript/issues/2503 .

`console.error` with a `try`/`catch` around it to see whats going on.

david-driscoll:
@nycdotnet, have any time to test it, I believe you worked on the original integration.

nycdotnet:
I did.  Ironically the very reason I did it was for the scope-per-iteration behavior of let in for loops which is now available in TS 1.7.  I don't have time to test it until after my Pluralsight course is done, but maybe in a week or two.  If it feels magical like "holy crap it worked" then it's probably fine.  That was my acceptance criteria.

basarat:
> BTW: loved your DNR episode.

:100: . Because of *that episode* I sneak peaked the syntactic highlight code (I have a new non atom version for [TypeScript](https://github.com/TypeScriptBuilder/tsb-docs#syntax-highlighting)) and will definitely be adding code lens (https://github.com/TypeScriptBuilder/tsb/issues/41 not hard to do thanks to CodeMirror's inline widget API)

david-driscoll:
I've been using this lately without any large issues, so I'm going to drop it in, and see if something gets reported.

nycdotnet:
Thanks man!

Nice work.  I'm glad you were able to promise-ify it as well.

As soon as my course is done, I'll be able to come up for air and check it out.

basarat:
Released : https://github.com/TypeStrong/atom-typescript/releases/tag/v7.12.0 :rose:

PR #729, opened 2015-11-17
Remove autocomplete from packages list.

kepler0:
Should autocomplete be listed if it is in core?


basarat:
> Should autocomplete be listed if it is in core?

Nope. Thanks. It wasn't in core once upon a time :rose:

Issue #682, opened 2015-10-22
Import pinkiepromise.d.ts


basarat:
To overcome tsconfig compile issue. (if you F6 the whole atom-typescript project you will see what I mean quite easily)

blakeembrey:
This is actually easy to fix with `typings` now :+1: Just `typings install npm:tsconfig -n tsconfig -S` (though a little verbose).

basarat:
@blakeembrey I actually got around it by just doing `declare module 'pinkie-promise'` locally : https://github.com/TypeStrong/atom-typescript/blob/1a3f638e0754c20f881744e88bc08dd92fa8b958/lib/typings/atompromise.d.ts#L100-L102

I have a seperate issue that I would love your input on, `npm run test` (just `ntsc -p ./lib`) succeeds locally but fails on travis : https://travis-ci.org/TypeStrong/atom-typescript I hear it has something to do with new npm flatdir : https://github.com/TypeStrong/atom-typescript/issues/733#issuecomment-158285850 I don't get it :) :rose:

blakeembrey:
Let me fix `tsconfig` here, I had to remove the `typings` field in `package.json` as it implied TypeScript could use it to resolve (which clearly was not the case, hence this issue). I can open a PR to use `typings`, if that makes sense.

basarat:
> I can open a PR to use typings, if that makes sense.

Sure ... PS: no hurry, love every bit of your effort :rose:

PR #738, opened 2015-11-23
Start using `typings`

blakeembrey:
I also removed all the `node_modules` - not sure what happened there? I also didn’t realise that there’s two `tsd.json` setups, so I only fixed the outer one (which wasn’t even one). I’ll patch the rest later, it’s a good exercise.

From #682


basarat:
Thank you so much :rose:

basarat:
> I also removed all the node_modules - not sure what happened there

initial laziness and unclarity if atom would install them or not

PR #747, opened 2015-11-25
catch `getEmitOutput` errors, cram them into errors list

### LINKED ISSUE #745: build never finishes (7.12.0)
I had problems like this in 7.11.2, but I attributed them to the "Progress events" problem that I was also having (and which seems to be somewhat ameliorated now).

I'm still not getting complete builds, though.
screen shot 2015-11-25 at 12 26 22 pm

The relevant tsconfig.json (unless it's finding some in the .Trash as well!) is

```
{
    "buildOnSave": false,
    "compileOnSave": true,
    "indentSize": 2,
    "tabSize": 2,
    "compilerOptions": {
        "target": "es6",
        "module": "commonjs",
        "experimentalDecorators": true,
        "emitDecoratorMetadata": true,
        "declaration": false,
        "noImplicitAny": false,
        "removeComments": false,
        "noLib": false,
        "preserveConstEnums": true,
        "suppressImplicitAnyIndexErrors": true,
        "inlineSources": true,
        "inlineSourceMap": true
    },
    "filesGlob": [
        "./typings/tsd.d.ts",
        "./global.d.ts",
        "./**/*.ts",
        "../gulpfile.js",
        "!**/bak/**"
    ],
    "files": [ "<omitted while pasting>" ],
    ],
    "atom": {
        "rewriteTsconfig": true
    }
}
```

I'd be interested in fixing this myself, or at least hunting it down, but I haven't yet grok'ed  where in the codebase the build process is managed.

The net results is that I can't do a build. Once it gets into this state (after the first build I try), it won't even try to do another unless I reload the atom view (in which case it will hang at this point the next time).

laurelnaiad:
This patch rather uncleanly handles errors thrown by `getEmitOutput` and pushes them into output.errors.

Two improvements are needed.

1. This isn't a very good solution to unhandled errors in the build. A cleaner and higher-level try/catch might make sense?
2. The underlying problem in my tsconfig.json could stand to be caught specifically and handled in a manner that helps the developer more than waiting for `getEmitOutput` to throw.

Closes #745

### DIFF
--- a/dist/main/atom/buildView.js
+++ b/dist/main/atom/buildView.js
@@ -25,8 +25,8 @@ function setBuildOutput(buildOutput) {
             mainPanelView.panelView.addBuild(new lineMessageView.LineMessageView({
                 goToLine: function (filePath, line, col) { return gotoHistory.gotoLine(filePath, line, col, gotoHistory.buildOutput); },
                 message: error.message,
-                line: error.startPos.line + 1,
-                col: error.startPos.col,
+                line: error.startPos ? error.startPos.line + 1 : null,
+                col: error.startPos ? error.startPos.col : null,
                 file: error.filePath,
                 preview: error.preview
             }));
--- a/dist/main/lang/modules/building.js
+++ b/dist/main/lang/modules/building.js
@@ -22,41 +22,62 @@ function diagnosticToTSError(diagnostic) {
 exports.diagnosticToTSError = diagnosticToTSError;
 function emitFile(proj, filePath) {
     var services = proj.languageService;
-    var output = services.getEmitOutput(filePath);
-    var emitDone = !output.emitSkipped;
     var errors = [];
-    var sourceFile = services.getSourceFile(filePath);
-    var allDiagnostics = services.getCompilerOptionsDiagnostics()
-        .concat(services.getSyntacticDiagnostics(filePath))
-        .concat(services.getSemanticDiagnostics(filePath));
-    allDiagnostics.forEach(function (diagnostic) {
-        if (!diagnostic.file)
-            return;
-        var startPosition = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start);
-        errors.push(diagnosticToTSError(diagnostic));
-    });
-    {
-        var sourceMapContents = {};
-        output.outputFiles.forEach(function (o) {
-            mkdirp.sync(path.dirname(o.name));
-            runExternalTranspiler(filePath, sourceFile.text, o, proj, sourceMapContents).then(function (additionalEmits) {
-                if (!sourceMapContents[o.name] && !proj.projectFile.project.compilerOptions.noEmit) {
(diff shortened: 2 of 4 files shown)


basarat:
Thanks a lot for this effort :rose: From https://github.com/TypeStrong/atom-typescript/issues/745#issuecomment-159755548

> 1) unhandled errors in the build shouldn't be lost! :)

They are not lost. They are intentionally logged to the console as unhandled, newly discovered edge cases.... like this one. But I guess  adding it to the *build* error instead of the default handling is good so :+1: 

> 2.) including .js files while not allowing nonTS extensions is a problem that ought not be caught only at the point where it throws an unhandled error.

Agreed. Fix belongs in `tsconfig.ts` See the other error scenarios : https://github.com/TypeStrong/atom-typescript/blob/80aee449b9cf25569ee1f067f2d86d5eaf9d4152/lib/main/tsconfig/tsconfig.ts#L175-L185 for samples :rose:

I'm happy to merge this PR if you think its what I should do :rose:

laurelnaiad:
> They are not lost. They are intentionally logged to the console as unhandled, newly discovered edge cases.... like this one.

... where nobody sees them unless the have the DevTools open. I'm not saying my PR is great or even good, but logging to DevTools console and **not** notifying the developer who is working on their own project in Atom kind of misses the point, no? 

I'm not going to win or lose by whether you merge this PR. I'd recommend that notifying the developer of any unhandled errors in atom-typscript, esp. if they're build-related, is critical, but I'm not in the best position to design how that will be done.

I'm really trying to crank out a proof-of-concept to use .ts in a project, so I need to get back to that. With the three PRs I've just splattered onto the repo, I'm now at a point where I feel like the tool is working, so I need to get back to using it instead of developing it. :)

If this were my repo and I were you, I'd amend Daphne's commit with improvements, since she doesn't know the codebase yet, then merge, as giving no feedback to a developer (unless they have DevTools open) on unhandled errors isn't a good thing. But it's up to you. I just think something needs to be done about notifying the developer on unhandled errors, since most developers working in Atom won't be sitting there with Atom's DevTools open inspecting the console for potential problems.

This wasn't even the first bug that slipped past me because I was just operating under the assumption that if there was a problem, I'd get a Growl and a line item in Build Errors.

As to addressing the specific issue of "to include .js files or not, and be consistent about it in tsconfig.json", thanks for pointing out where the right place is. I had a feeling you'd point to that file. I don't think that issue is nearly as critical as is the subject of ensuring unhandled errors bubble up into Atom GUI and not just the DevTools console.

Anyway, it's up to you whether to merge, of course. I have a feeling that with a little work, you'd be able to do what I was trying to do in a way that's more consistent with the overall architecture and design intents of the package. If you don't have the time to do that (and since I'm begging off for a little while to do my "real job"), then perhaps merging my commit makes sense? I don't know. :)

basarat:
I've created a clearer issue : https://github.com/TypeStrong/atom-typescript/issues/752 Closing as this is not the long term fix :rose:

PR #750, opened 2015-11-26
use externalTranspiler before doing diffs for status

### LINKED ISSUE #736: JS Outdated by opening any .ts file in editor
Hi,

I'm pretty new to typescript and atom-typescript, so this could be user error.

Every time I open a .ts file in the editor, a  **JS Outdated** message appears.

`~tsconfig.json JS Outdated`

Every time I save my .ts file, the JS Outdated is replaced with a happy green JS Current message (with no build errors) until I open some other .ts file.

*(Aside: ~tsconfig.json is in gray, JS Outdated in red -- do those two snippets have anything to do with each other? Or is the ~tsconfig.json just telling me which tsconfig.json is effective for for the .ts file I have open? I hope it's the latter, as if I can get atom-typescript to work I'll be using it on project which needs to have three distinct tsconfig.json configurations).*

I have verified that:

* the .ts file has not been modified on disk
* the .ts file has not been modified in buffer
* the tsconfig.json isn't even in a buffer and hasn't been modified

So basically, just looking at a .ts file in atom is making it *seem* to atom-typescript as if it's outdated.

This is incredibly distracting and confusing, as well as rendering whatever value I would be be getting from the message unavailable to me. Of course, I also wonder if it's somehow trying to tell me that something is wrong with my configuration and it's trying to fix it.

Am I "doing it wrong"?

tsconfig.json:

```
{
  "externalTranspiler": "babel",
  "buildOnSave": false,
  "compileOnSave": true,
  "indentSize": 2,
  "tabSize": 2,
  "compilerOptions": {
    "declaration": true,
    "allowNonTsExtensions": false,
    "module": "commonjs",
    "target": "es6",
    "noImplicitAny": false,
    "removeComments": false,
    "sourceMap": true,
    "noLib": false,
    "preserveConstEnums": true,
    "suppressImplicitAnyIndexErrors": true
  },
  "filesGlob": [
    "./*.ts",
    "!./node_modules/**/*",
    "./env/**/*.ts"
  ],
  "files": [
    "./global.d.ts",
    "./gulpfile.d.ts",
    "./gulpfile.ts",
    "./env/index.d.ts",
    "./env/index.ts",
    "./env/processMgmt/index.d.ts",
    "./env/processMgmt/index.ts"
  ],
  "atom": {
    "rewriteTsconfig": true
  }
}
```

I have mucked around with this config to try to figure out if one or the other flag is behind this behavior to no avail (yet).

* tsc --version: message TS6029: Version 1.6.2
* tsd --version: >> tsd 0.6.5
* atom --version: 1.2.3
* what other software is/could be involved? Since I can see that the buffer is unmodified, I don't expect this has to do with some other package being evil and going in and changing source code...

~~P.S. While I'm at it, and just in case it's related, I also get a little Github-flavored kitty icon in the lower right of Atom listing a series of Pending Requests, and I have my suspicions that atom-typescript is related to these pending requests. I can quite atom and restart it and these pending requests will eventually show up again sooner or later.... potentially a related issue? If not, what's the easy way to get these requests to complete?~~ (EDIT: I see this has already been diagnosed as "some other atom problem"...)

### LINKED ISSUE #737: JS Status is intially misleading
Its starts out with `JS Outdated` 

image

It should really be something like `Inderterminite` till we do an actual validation. 

## Current

* Its `oudated` by default < This needs to change
* Its `current` if user saves file and `compileOnSave` was true (or user does a build)
* Its `outdated` if the user edits the file after the last time we did a compile for a file

refs https://github.com/TypeStrong/atom-typescript/issues/736

laurelnaiad:
Create new function `getRawOutputPostExternal` with same interface
as getRawOutput except that it returns a <Promise<ts.EmitOutput>
instead of a ts.EmitOutput.

This function calls getRawOutut, then passes the result through
`runExternalTranspiler` so that babel, if configured, is applied.

Call/use this new function in `getOutputJsStatus`.

Net result: people using babel are not constantly being told their
JS is Outdated.

Closes #737

### DIFF
--- a/dist/main/lang/modules/building.js
+++ b/dist/main/lang/modules/building.js
@@ -78,6 +78,18 @@ function getRawOutput(proj, filePath) {
     return output;
 }
 exports.getRawOutput = getRawOutput;
+function getRawOutputPostExternal(proj, filePath) {
+    var output1 = getRawOutput(proj, filePath);
+    var sourceFile = proj.languageService.getSourceFile(filePath);
+    var sourceMapContents = {};
+    return runExternalTranspiler(filePath, sourceFile.text, output1.outputFiles[0], proj, sourceMapContents).then(function () {
+        return {
+            outputFiles: output1.outputFiles,
+            emitSkipped: false
+        };
+    });
+}
+exports.getRawOutputPostExternal = getRawOutputPostExternal;
 function getBabelInstance(projectDirectory) {
     return new Promise(function (resolve) {
         if (!babels[projectDirectory]) {
--- a/dist/main/lang/projectService.js
+++ b/dist/main/lang/projectService.js
@@ -599,23 +599,27 @@ exports.getOutputJs = getOutputJs;
 function getOutputJsStatus(query) {
     projectCache_1.consistentPath(query);
     var project = projectCache_1.getOrCreateProject(query.filePath);
-    var output = building_1.getRawOutput(project, query.filePath);
-    if (output.emitSkipped) {
-        if (output.outputFiles && output.outputFiles.length === 1) {
-            if (output.outputFiles[0].text === building.Not_In_Context) {
+    return building_1.getRawOutputPostExternal(project, query.filePath)
+        .then(function (output) {
+        if (output.emitSkipped) {
+            if (output.outputFiles && output.outputFiles.length === 1) {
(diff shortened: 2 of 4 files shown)


laurelnaiad:
Word to the wise: github's new gui has bugs....  don't trust it to put the most up-to-date commit messages into the PRs it generates for you. :)

PR #751, opened 2015-11-26
use externalTranspiler before doing diffs for status

### LINKED ISSUE #737: JS Status is intially misleading

laurelnaiad:
Create new function `getRawOutputPostExternal` with same interface
as getRawOutput except that it returns a `<Promise<ts.EmitOutput>`
instead of a ts.EmitOutput.

This function calls getRawOutut, then passes the result through
`runExternalTranspiler` so that babel, if configured, is applied.

Call/use this new function in `getOutputJsStatus`.

Net result: people using babel are not constantly being told their
JS is Outdated.

### DIFF
--- a/lib/main/lang/projectService.ts
+++ b/lib/main/lang/projectService.ts
@@ -124,7 +124,7 @@ export function build(query: BuildQuery): Promise<BuildResponse> {
         return output;
     });
 
-    // Also optionally emit a root dts:		
+    // Also optionally emit a root dts:
     building.emitDts(proj);
 
     // If there is a post build script to run ... run it
@@ -138,7 +138,7 @@ export function build(query: BuildQuery): Promise<BuildResponse> {
             }
         });
     }
-    
+
     let tsFilesWithInvalidEmit = outputs
         .filter((o) => o.emitError)
         .map((o) => o.sourceFileName);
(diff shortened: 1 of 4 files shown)


lib/main/lang/projectService.ts
```diff
@@ -938,22 +938,26 @@ interface GetOutputJsStatusResponse {
 export function getOutputJsStatus(query: FilePathQuery): Promise<GetOutputJsStatusResponse> {
     consistentPath(query);
     var project = getOrCreateProject(query.filePath);
-    var output = getRawOutput(project, query.filePath);
-    if (output.emitSkipped) {
-        if (output.outputFiles && output.outputFiles.length === 1) {
-            if (output.outputFiles[0].text === building.Not_In_Context) {
+    return getRawOutputPostExternal(project, query.filePath)
```
basarat: Think it should only be called if the project config says that an external transpiler needs to be used :rose: or am I misreading

laurelnaiad: `runExternalTranspiler` (which is called by `getRawOutputPostExternal`) is kind of a misnomer. It's a no-op if an external transpiler isn't configured to run. Ref: https://github.com/laurelnaiad/atom-typescript/blob/iss737/lib/main/lang/modules/building.ts#L168

basarat:
Cheers. Will publish with next release :rose:

basarat:
Released : https://github.com/TypeStrong/atom-typescript/releases/tag/v7.13.0 :rose:

basarat:
Reversing this commit. Just because I'm lazy, its 9pm on a Friday :) and this causes https://github.com/TypeStrong/atom-typescript/issues/765 :rose:

Issue #767, opened 2015-12-05
Support dot notation in jsx


unional:
````
var x = {
  y: <Tab title='Nice Tab' display.bind='...' > // Error on `display.bind=...` because of the dot
}
````

`display.bind` is aurelia-like syntax, JSX (thus TSX) does not seems to object that notation.

Issue #675, opened 2015-10-21
Reuse Grammars from Sublime-TypeScript


blakeembrey:
From https://github.com/TypeStrong/atom-typescript/pull/672, we should investigate if a simple npm script can be utilized to copy and repackage the grammars from Sublime.

## edits by @basarat: 

We have very little custom work ... 

* The top that needs to be preserved across versions: https://github.com/TypeStrong/atom-typescript/blob/729b5522cc17fc84511c05ebb5ae1812f3ec4b00/grammars/typescript.cson#L12-L80

* The angular template support
https://github.com/TypeStrong/atom-typescript/pull/798

stepancar:
@blakeembrey, I think we should ask typescript team. Maybe they have any plans to export/support typescript grammar as any external module.
Can you [ask](https://github.com/Microsoft/TypeScript/issues), please? I don't know how to ask correctly.
Way when we just copy grammars from sublime is not good, i think

blakeembrey:
@stepancar I think it's worth investigating from the Atom side first too. If we can just convert it with a script, there's no need for a separate module or special support. As for them publishing an external module, I don't have any ideas on the format of how that would look - I haven't had enough exposure to various IDE language definitions to know if it's simple to publish a common denominator (or maybe generate each of them?).

stepancar:
@blakeembrey i created [issue](https://github.com/Microsoft/TypeScript/issues/5356)
join to discussion :)

blakeembrey:
:tada: It's done. We should be investigating this soon, would be easy enough to use a script to convert into the format Atom uses. It looks like they're going to add a testing framework too, so that sounds like a huge win against maintaining the grammar here.

sgal:
@blakeembrey As far as I see, those grammar files are just imported from Sublime Text package, so it's not a "real" TS grammar from MS. See https://github.com/Microsoft/TypeScript-TmLanguage/issues/1

We indeed should track the progress there, but I guess it's too early to integrate it here.

quadrophobiac:
@blakeembrey I arrived here after indicating interest in one of the separate tickets (#707) prior to it's collation here by the recent mass assignment of 'up-for-grabs' labels. I read from your comment https://github.com/TypeStrong/atom-typescript/issues/675#issuecomment-154325788 that you are following the 'copy via script' rather than 'investigating from Atom first principles' option on this - did I understand that correctly? Does that mean that these are or are not still 'up-for-grabs' style issues?

[...]

PR #770, opened 2015-12-06
Support `a.b` as attribute name

unional:
Fix https://github.com/TypeStrong/atom-typescript/issues/767


basarat:
I've merged ... just because I don't see any harm 

This work might get lost when we do : https://github.com/TypeStrong/atom-typescript/issues/675 though :rose:

unional:
Will file PR again if that happens. :+1:

PR #777, opened 2015-12-12
Adding an external resolver in tsconfig.json

### LINKED ISSUE #776: Add support for projects built with WebPack
In WebPack the imports are resolved based on the aliases in the configuration file. The path system is different than the one in TypeScript.

Add a method to support external resolvers, like webpack.

achicu:
I've added a new property in `tsconfig.json` called `externalResolver`. The property can be either a string or a `{ name : string, options: any }`.

This fixes #776 

The name is a relative path to the folder of the `tsconfig.json` file. The external resolver should be a JavaScript file that exports a constructor of the resolver (see `ResolverConstructor` below).

    export interface ResolvedModule {
        resolvedFileName: string;
        /*
         * Denotes if 'resolvedFileName' is isExternalLibraryImport and thus should be proper external module:
         * - be a .d.ts file
         * - use top level imports\exports
         * - don't use tripleslash references
         */
        isExternalLibraryImport?: boolean;
    }

    interface ResolverInstance {
        resolveModuleNames(moduleNames: string[], containingFile: string): ResolvedModule[];
    };
    
    type ResolverConstructor = new (config: tsconfig.TypeScriptProjectFileDetails, options? : any) => ResolverInstance;

This is an example `tsconfig.json`

    {
        "compilerOptions": {
            "module": "commonjs",
            "noImplicitAny": true,
            "sourceMap": true,
            "target": "ES5"
        },
        "compileOnSave": false,
        "buildOnSave": false,
        "externalResolver": "resolver.js"
    }

And example `resolver.js` file.

    var webpack = require("webpack");
    var path = require("path");
    var fs = require("fs");
    var makeResolver = require("awesome-typescript-loader/dist/resolver");

    function WebPackResolver(config, options) {
    this.options = options || { config: "./webpack.config.js" };
    this.makeResolver();
    }

    WebPackResolver.prototype.makeResolver = function () {
        this.resolver = null;
        if (this.watcher) {
            this.watcher.close();
        }

        var configPath = path.join(__dirname, this.options.config);
        var self = this;
        this.watcher = fs.watch(configPath, { persistent: false, }, function () {  self.makeResolver(); });

        var config;
        try {
            delete require.cache[configPath];
            config = require(configPath);
        } catch (err) {
            console.error(err);
        }

        try {
            this.resolver = makeResolver.default(webpack(config || { }).options);
        } catch (err) {
            console.error(err);
        }
    }

    WebPackResolver.prototype.resolveModuleNames = function (moduleNames, containingFile) {
        var resolver = this.resolver;
        if (!resolver) {
            return [];
        }

        return moduleNames.map(function (moduleName) {
            var resolvedFileName;
            try {
                resolvedFileName = resolver.resolveSync(path.dirname(containingFile), moduleName);
            } catch (err) {
                console.error(err);
                return;
            }
            return { resolvedFileName: resolvedFileName }
        });
    }

    module.exports = WebPackResolver;

### DIFF
--- a/lib/main/lang/core/languageServiceHost2.ts
+++ b/lib/main/lang/core/languageServiceHost2.ts
@@ -209,6 +209,10 @@ export var getDefaultLibFilePath = (options: ts.CompilerOptions) => {
 
 export var typescriptDirectory = path.dirname(require.resolve('ntypescript')).split('\\').join('/');
 
+interface ResolverInstance {
+    resolveModuleNames(moduleNames: string[], containingFile: string): ts.ResolvedModule[];
+};
+type ResolverConstructor = new (config: tsconfig.TypeScriptProjectFileDetails, options? : any) => ResolverInstance;
 
 // NOTES:
 // * fileName is * always * the absolute path to the file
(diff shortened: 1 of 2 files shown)


laurelnaiad:
This might also help with systemjs projects. Hopefully it would be a short-term fix until Typescript 1.8 comes out which will hopefully have its own mechanism for this.

basarat:
Sorry. I should have looked at this sooner. As much as I like the intention and the effort (:+100:) I am averse to adding features I don't use (would rather have the TS team solve it https://github.com/Microsoft/TypeScript/issues/5039) as it increases the maintainability burden. I am looking to simplify this project as I want it to do into sleep + stable mode :rose:

PR #787, opened 2015-12-17
add systemjs import syntax support to quick fix

jschwarty:
Couple of thoughts on possible changes to this PR:

- I used the project file compiler options to identify if that is set to use "system" in order to make the decision on the syntax for the import statement. Not sure if this is ideal or not...
- I hard coded double quotes similar to the current statement, but could this be changed so the single/double quote setting from atom would be used here?

### DIFF
--- a/dist/main/lang/fixmyts/quickFixRegistry.js
+++ b/dist/main/lang/fixmyts/quickFixRegistry.js
@@ -1,6 +1,7 @@
 "use strict";
 var addClassMember_1 = require("./quickFixes/addClassMember");
 var addClassMethod_1 = require("./quickFixes/addClassMethod");
+var addImportFromStatement_1 = require("./quickFixes/addImportFromStatement");
 var addImportStatement_1 = require("./quickFixes/addImportStatement");
 var equalsToEquals_1 = require("./quickFixes/equalsToEquals");
 var extractVariable_1 = require("./quickFixes/extractVariable");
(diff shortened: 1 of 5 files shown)


jschwarty:
I am thinking that the `tsconfig.json` compiler module setting should not be the deciding factor because in Angular2 the compiler options need to be module "commonJs", so maybe a package setting that allows a user to configure the import syntax?

basarat:
@jschwarty thanks Justin! :rose: 

> I used the project file compiler options to identify if that is set to use "system" in order to make the decision on the syntax for the import statement. Not sure if this is ideal or not.

Not. Please create a *new* quickfix entry instead of replacing the current one. Let the user decide. Copy paste into a new file is fine :rose:

> but could this be changed so the single/double quote setting from atom would be used here

Not easily as the code is running in the *backend* worker that has limited `atom` API access so might not work or break in the future. I wouldn't do it :rose:

jschwarty:
Ok, I will be afk for the weekend so I will make these changes at the start of next week and update the PR. Thanks!

nycdotnet:
Thanks Schwarty!

jschwarty:
Created a new PR (https://github.com/TypeStrong/atom-typescript/pull/791) with the suggested approach...

amb-it-ion:
I got a missing dependency "Cannot find name <xyz>", but get no quick fix suggestion. I am on version 7.16.1. Any advice?

Thanks for your work - this feature is much needed.

PR #798, opened 2015-12-30
Angular2 Decorator Template Highlighting

jbarket:
Added simple support for Angular2 inline template highlighting. In short, if we find something that looks like this:

@Component({
  template: `
    <some><html><here></here></html></some>
  `
})

We activate text.html.basic highlighting on the contents of the multiline string.

It's possible that we could look specifically for @Component, but right now, we're piggybacking off of #decorator to prevent weirdness/code duplication.

To make this work, I had to remove the $base include from #decorator. If people begin passing anonymous functions and so on to decorators, this will need revisiting. But right now, we're talking about booleans and objects, so I don't know that it really needs $base. If it does, probably better to just catch @Component->template on its own.

### DIFF
--- a/grammars/typescript.cson
+++ b/grammars/typescript.cson
@@ -1383,13 +1383,7 @@
 
   'angulartemplate':
     'begin': '\\b(template:\\s*`)'
-    'beginCaptures':
-      '0':
-        'name': 'storage.type.annotation.js'
     'end': '(`)'
-    'endCaptures':
-      '0':
-        'name': 'storage.type.annotation.js'
     'patterns': [
       { 'include': 'text.html.basic' }
     ]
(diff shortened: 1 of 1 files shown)


basarat:
Thanks. Work with https://github.com/TypeStrong/atom-typescript/issues/675 should preserve this :rose:

basarat:
released : https://github.com/TypeStrong/atom-typescript/releases/tag/v7.16.0 :rose:

ghpabs:
Thanks guys! I was looking forward to this feature. :clap:

basarat:
Note that we are about to get automated upstream grammar : https://github.com/TypeStrong/atom-typescript/pull/813 That will lose this work. A PR modifying `grammar.js` automation script to support this would be appreciated once that is merged :rose:

alindsay55661:
So, how do I active this? I am using version 8.1.0 and syntax highlighting isn't available for template strings.

Issue #810, opened 2016-01-11
Auto-Format on save


lordmat0:
Hi there,

Just wondering if this was already included or if there is a way to do this? If not I'll give it a go to include it.

Thanks
Mat

basarat:
> Just wondering if this was already included or if there is a way to do this? If not I'll give it a go to include it.

No included. If you want to add it please add under the `atom` key in tsconfig :rose: https://github.com/TypeStrong/atom-typescript/blob/master/docs/tsconfig.md#atom

lordmat0:
Okay I'll give it a go tonight :)

basarat:
Fix https://github.com/TypeStrong/atom-typescript/pull/815 :rose:

basarat:
Released : https://github.com/TypeStrong/atom-typescript/releases/tag/v8.1.0 :rose:

PR #815, opened 2016-01-12
Auto format on save

lordmat0:
#810 

I wasn't sure if this should be a default on, so I decided to have it defaulted to off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ JavaScript developers can now just open a `.ts` file and start hacking away like
 * Project Build Support
 * `package.json` Support
 * React Support
-* Format code
+* Format code (configurable to be on save)
 * Goto Declaration
 * Find References
 * Block comment and uncomment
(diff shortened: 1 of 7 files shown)


basarat:
> so I decided to have it defaulted to off.

Definitely. Don't want user to open someone else's project's typescript file for quick editing just to have it change on them :rose:

basarat:
Please answer: Can you try on a project that doesn't have the `atom` key already. Does it work (the tsconfig.json shouldn't be modifed and format on save shouldn't happen and there should be no errors in the atom debug console)? :rose:

lordmat0:
I tried it already, it works correctly :)

We create the key here:
https://github.com/lordmat0/atom-typescript/blob/auto_format_on_save/lib/main/tsconfig/tsconfig.ts#L402-L406

Before I use it here:
https://github.com/lordmat0/atom-typescript/blob/auto_format_on_save/lib/main/tsconfig/tsconfig.ts#L445



Would it also not be worth adding it as a menu item?

basarat:
> Would it also not be worth adding it as a menu item? so others using atom on the same project don't have to share the same setting.

You mean *atom config* item? No. Its good to have this shared as a requirement on using the project :rose:

basarat:
Released : https://github.com/TypeStrong/atom-typescript/releases/tag/v8.1.0 :rose:

s00500:
Has this feature been removed again? Why?

PR #831, opened 2016-02-02
Fix tsconfig.md regarding ./ prefix

PhilipMay:
Here is a little fix for the tsconfig.md documentation.
You mention a 'gs' snippet but the example in tsconfig.md differs from the snippet.
Is does not just differ but it seems to have a bug. I tried it and
the ./ prefix does not work for me - so I removed it for you.
Also the generated files section has no ./ prefix.

### DIFF
--- a/docs/tsconfig.md
+++ b/docs/tsconfig.md
@@ -59,7 +59,7 @@ A default `filesGlob` is available for you as a snippet : `fg`
     },
     "filesGlob": [
         "**/*.ts",
-        "!node_modules/**/*.ts"
+        "!node_modules/**"
     ],
     "files": [
         "globals.ts",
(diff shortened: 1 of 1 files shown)


basarat:
You are right. Actually can you add `tsx` as well. For reference this is the actual ideal glob : https://github.com/TypeStrong/atom-typescript/blob/f50fe08e7b3d1b33d01f6a02b9319e488bc97fbb/lib/main/tsconfig/tsconfig.ts#L224-L226 :rose:

Please leave a comment once done :+1:

PhilipMay:
> Please leave a comment once done :+1: 
Done 👍🏽

basarat:
Thank you :rose: :heart:

Issue #833, opened 2016-02-03
Add support for path mapping to `tsconfig.json`.


mko:
Recently (about a week ago), TypeScript added support for a new `moduleResolution` type called Path Mapping, which allows the user to specify a `baseUrl` object and `paths` object for mapping module names to their actual paths, allowing users to not be tied to only using `node` (i.e. NPM) as an alternative module resolution mode. This opens up the ability for devs to use alternative package managers (like JSPM) and using simplified module import statements for local modules as well.

Path mapping is enabled by setting up your `tsconfig.json` like so:

``` json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "angular2/*": ["../path/to/angular2/*"],
      "local/*": ["../path/to/local/modules/*"]
    }
  }
}
```

and using the imports in TypeScript files like so:

``` ts
import { bootstrap } from 'angular2/bootstrap';
import { module } from 'local/module';
```

This provides a fix for working with client-side modules installed via [JSPM](http://jspm.io). You can see the details of the implementation in the PR 5728 on the TypeScript project: https://github.com/Microsoft/TypeScript/pull/5728/files

The TypeScript nightly has the fix in place, though the fix isn't in the stable version on NPM, so you have to install TypeScript by `npm install -g typescript@next` instead of just `typescript` to reproduce this issue.

Before Path Mapping, I saw numerous errors related to the JSPM module name mappings:

![Before Path Mapping](https://cloud.githubusercontent.com/assets/85261/12799721/366680f6-ca85-11e5-9247-89aae0ec8784.png)

After Path Mapping, those errors disappear and now I see the error about `baseUrl` being an invalid value: 

![After Path Mapping](https://cloud.githubusercontent.com/assets/85261/12799727/48c36cbe-ca85-11e5-9d78-8523affcdeac.png)

Clearly, this is a "bleeding edge" issue only, but I know that many other devs using both Angular2 and JSPM are running into this issue. A fix would make the experience of developing Angular2 with JSPM in Atom much more pleasant.

basarat:
btw nice user handle :+1: @mko :rose:

laurelnaiad:
@mko , thank you for pushing this along. Are you saying that you changed your tsconfig, wrote some mappings and it "just worked" in atom-typescript (except for the complaint about the tsconfig schema)?

If so, that's great! This feature isn't only useful for angular2 apps, though they are, of course, a significant contingent among the stakeholders.

[...]

PR #842, opened 2016-02-09
#833 Add support for path mapping in `tsconfig.json`

mko:
Note: In order for the path mapping to be successful, you must be using `typescript@next` as of February 9th, 2016.

It might be worth holding on to this PR until TypeScript has fully documented this feature, but I wanted to make sure others who might encounter this issue trying to use path mapping were able to fix it themselves if needed.

Additionally, I haven't had time to fully test this yet like I would like, but it cleared up the errors in Atom and the compilation appears to be working properly again.


mko:
Resolves #833 by allowing for `baseUrl` and `paths` in the `tsconfig.json` validator.

mko:
Scratch this PR for now. After further testing, it ended up revealing that getting path mapping working with `atom-typescript` is going to take more work than just enabling the new `tsconfig.json` objects.

nycdotnet:
Thanks for submitting this regardless.  Hope we'll be able to support this soon.

PR #861, opened 2016-02-24
Support for option "disableCompileOnSave", which is used by atom editor

kerbyfc:
Atom use "disableCompileOnSave" to prevent single file compilation on save


kerbyfc:
I'm addict, sorry

PR #862, opened 2016-02-24
Updated typescript version, to pickup additions

david-driscoll:
@blakeembrey @nycdotnet @Cu3PO42 (and anyone else)

Think we can push a release through at all with the updated services, running into a few bugs that have been fixed by the latest TS releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "immutable": "^3.7.3",
     "json2dts": "0.0.1",
     "mkdirp": "^0.5.0",
-    "ntypescript": "1.201601272211.1",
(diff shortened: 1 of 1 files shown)


blakeembrey:
I think we want to stick to pinning releases. Things could break randomly in the future otherwise.

david-driscoll:
whoops, good 'ol tool that updates the values, updated the PR.

blakeembrey:
I'll look at doing a release soon :+1:

david-driscoll:
rock on... I'll try and make some time to work on the plugin a little bit more, help solve some of the issues I run into, that I imagine others do too. :smile:

PR #889, opened 2016-03-17
tsconfig.json is saved with the same new line characters as the previous one

### LINKED ISSUE #678: tsconfig.json always save to CRLF on Windows environment
Atom Editor now has line ending selector for selecting ending format of file.

But on Windows environment even I try to select LF format for tsconfig.json which always be saved to CRLF format.

Does it have limitation to solve the issue? Thanks a lot!

kondei:
I fixed #678 to be able to save tsconfig.json with LF in Windows.
Used https://github.com/sindresorhus/detect-newline because there are already used detect-indent.


basarat:
Cheers :rose:

Issue #890, opened 2016-03-18
Remove `Promise.defer`


basarat:
Its removed from latest chrome releases

PR #891, opened 2016-03-18
Promise.defer no more

basarat:
closes https://github.com/TypeStrong/atom-typescript/issues/890


basarat:
PS: the test infrastructure needs work. Its broken at the moment as I get a ton of duplicate ident errors locally `npm run build` but the build server says "all good" :rose:

basarat:
Marking this as a `minor` as its a significant change ... but should work just fine :rose:

PR #900, opened 2016-03-30
Fix String Highlighting

joshheyse:
I noticed that strings were no longer being colored correctly.  This may have been due to a recent change where the styles of Atom now include quoted in the name.

I figured this out by comparing the styles between atom-typescript and the other language syntax plugins.

### DIFF
--- a/grammars/ts.cson
+++ b/grammars/ts.cson
@@ -602,7 +602,7 @@ repository:
           "2":
             name: "keyword.other.ts"
           "3":
-            name: "es6import.path.string"
+            name: "es6import.path.quoted.string"
       }
       {
         comment: "Match import = require"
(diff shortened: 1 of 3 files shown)


blakeembrey:
Those are auto-generated, please apply any grammar changes to https://github.com/TypeStrong/atom-typescript/blob/master/scripts/grammar.js

joshheyse:
Thanks Blake. I missed that they were generated and I'll fix and resubmit

basarat:
@joshheyse its been working fine without `quoted`: 

image

What is the theme you are using (I'm on monokai)? :rose:

joshheyse:
I'm using the Solarized Dark theme.  I compared the classes to the js/jsx syntax highlighting which is where i found the addition of the quoted class.

ts/tsx

screen shot 2016-04-07 at 21 16 43

js/jsx

screen shot 2016-04-07 at 21 17 06

basarat:
Thanks :rose:

basarat:
https://github.com/TypeStrong/atom-typescript/releases/tag/v8.6.0 :rose:

Issue #888, opened 2016-03-17
Project file contains invalid options forceConsistentCasingInFileNames


denis-sokolov:
A [compiler option](https://github.com/Microsoft/TypeScript/wiki/Compiler-Options) forceConsistentCasingInFileNames causes the entire atom-typescript to fail. No typechecks or suggestions of any sort work when the option is in the config.

```
Error: Project file contains invalid options
    at Object.getProjectSync (~/.atom/packages/atom-typescript/dist/main/tsconfig/tsconfig.js:260:32)
```

More generally, perhaps you should not handle unknown keys gracefully?

princjef:
I run into the same problem using the same compiler option.

While I do appreciate a warning in the case of unknown compiler options (handy in case of misspellings and such), it seems like they should just be ignored for static analysis or there should at least be an option to ignore them.

wiggisser:
I have the same problem with the option "listEmittedFiles".

PR #930, opened 2016-04-20
Update compiler options validation for tsconfig

princjef:
Addresses the surface-level issue from #888, namely that there are some new compiler options from TypeScript that are not included in the tsconfig validation. The following have been added (pulled from [tsconfig.json schema](http://json.schemastore.org/tsconfig):

 * forceConsistentCasingInFileNames
 * pretty
 * project
 * reactNamespace

There is a deeper question from the issue regarding whether unrecognized tsconfig options should result in an error (rather than a warning), but I will leave that for a separate discussion/PR.


basarat:
:rose:

basarat:
https://github.com/TypeStrong/atom-typescript/releases/tag/v8.9.1

Issue #923, opened 2016-04-13
custom `typescriptServices` still references `lib.d.ts` from `ntypescript`


leonyu:
After setting a custom `typescriptServices`, `atom-typescript` still references `lib.d.ts` from `ntypescript`. 

This causes incompatibility when the versions are significantly different.

pcan:
Any news about this? Visual Studio allows this kind of customization, see [here](https://github.com/Microsoft/TypeScript/wiki/Dev-Mode-in-Visual-Studio#using-a-custom-language-service-file). Another inputbox in the settings pane (like the one for `typescriptServices`) should be enough, right?

PR #951, opened 2016-05-15
Load lib.d.ts from typescriptServices path

seanchas116:
This is a fix for #923, to load default lib from custom `typescriptServices` path.


basarat:
Released with `8.10.2` minor version :rose: 

If its cool with everyone I'll merge and publish if it looks good enough to keep my effort at a minimum without letting the code go toooo stale :rose:

Issue #726, opened 2015-11-15
Keyboard shortcut for Type information at cursor


basarat:
Currently only shown on Mouse Hover : https://github.com/TypeStrong/atom-typescript#type-information-on-hover

Requested by @JaKXz here https://github.com/TypeStrong/atom-typescript/issues/716#issuecomment-156810461

daniel-trinh:
+1

michalczukm:
Hi, I would like to grab this issue. Should I get somehow assigned on issue? contributing.md doesn't cover that :smile: 

PS Any concrete key-binding suggested?
R# / JetBrains IDEs have ctrl+q for similar functionality. I would suggest to use that.

nycdotnet:
Go ahead and implement. If ctrl+q isn't used already with Atom then it sounds good.  :)

michalczukm:
ok, thanks for answer :)

dylanscott:
Hope I'm not stepping on anyones toes, but I was sorely missing this feature after switching to Atom recently so I took a crack at it with #970

basarat:
Released with `8.11.0` (minor update). Thanks @dylanscott :rose:

PR #970, opened 2016-06-06
#726 adds show-type command for type info at cursor

dylanscott:
This PR adds a `show-type` command that displays type information at the cursor. The overlay looks like this:
screen shot 2016-06-05 at 10 22 35 pm

Some notes:
* The type information overlay is dismissed when the cursor position changes, the escape key is pressed, or when the editor the overlay is in loses focus.
* I did not merge this code path with the one that displays type information on hover, mostly out of laziness.
* This is my first change on an Atom package so any improvement suggestions would be very welcome.


basarat:
Looks good to me and I don't want to block you. Thanks! :heart: 

If anybody reports a bug, I'll just revert and ping you :rose:

basarat:
Released with `8.11.0` :rose: Thanks again :heart:

dylanscott:
Awesome, thanks for the quick turnaround!

Issue #448, opened 2015-07-09
Implement hyperclick package


park9140:
Part of Nuclide is the hyperclick package, https://atom.io/packages/hyperclick, it allows for interacting with text in the editor while holding the alt key, via a service.  

Some thoughts of things we could implement with this.
- Jump to definition.
- allow quick fixes to define syntax errors they are associated with
  - for example quick fix to implement an interface could be attached to the, the syntax error for a interface not being implemented.

adidahiya:
+1 this would be awesome!

littleStudent:
Is anyone working on this? would love to see this feature added to atom-typescript.
Never did atom package development, would be a lot work for me to get into this stuff.

basarat:
> Is anyone working on this?

Nope :rose:

cevek:
+1

marcoskichel:
+1

markvandenbrink:
This is very nice, but there is a problem using the `ctrl`-key, since this key is used to make multiple selections/use multiple text cursors in Atom (which is very useful). Maybe it is better to use the `alt`-key as described in the readme of hyperclick:

> - `<alt>` or `<alt-mousemove>` underlines clickable text under the mouse.
> - `<alt-mousedown>` performs the action associated with the clickable text.
> - `<cmd-alt-enter>` performs the action on the text under the cursor.

wingyplus:
@markogresak Atom Typescript implement hyperclick provider for jump to declaration since version 10.0.1 . You can use `alt` key that hyperclick mapped.

markvandenbrink:
It seems the ctrl-key is mapped on Windows instead of the alt-key. Submitted a PR to change this behavior: https://github.com/facebook/nuclide/pull/612

tlaverdure:
I agree with @markvandenbrink about the keymap issue. I'm used to using **cmd + click** to make multiple selections. 

Is this also only supposed to work with typescript files? Seem to see the white hyperclick underline in every file type I'm in when holding down **cmd** which makes this feature buggy.

robofish:
If i'm not mistaken, the integration is gone again with atom typescript v11? 😅

guncha:
[...]

PR #995, opened 2016-07-06
Implement go to declaration with hyperclick provider

wingyplus:
Hi 

I see #448 talk about this.​ Any suggestions are welcome.

### DIFF
--- a/lib/hyperclickProvider.ts
+++ b/lib/hyperclickProvider.ts
@@ -6,6 +6,8 @@ const TS_GRAMMARS = Set<string>(["source.ts", "source.tsx"]);
 
 export let providerName = "typescript-hyperclick-provider";
 
+export let wordRegExp = /([A-Za-z0-9])+|['"`](\\.|[^'"`\\\\])*['"`]/g;
+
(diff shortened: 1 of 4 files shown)


basarat:
thanks @wingyplus :rose: :heart: https://github.com/TypeStrong/atom-typescript/releases/tag/10.0.0

wingyplus:
@basarat It doesn't work. Is it compile manually?

basarat:
> Is it compile manually?

sadly. As atom packages are essentially just `git tags` so whatever is in git is what makes it to the user system (other than `npm install`) :rose:

tlaverdure:
Should this only work for typescript files? It seems to be activation on all file types like css, less, scss, html...

athyuttamre:
This is a big step up! Thanks for implementing!

wingyplus:
@tlaverdure hyperclick enabled all filetypes. but in feature "go to declaration" work only typescript filetypes (`.ts` and `.tsx`).

tlaverdure:
Maybe using the alt key would be better so it doesn't conflict with default atom functionality.

PR #1012, opened 2016-07-15
feat(tsconfig): add lib option support

heruan:


basarat:
`10.1.4` :rose:

PR #1015, opened 2016-07-17
Support TS 2.0 flag noUnusedParameters

spiffytech:


basarat:
Thanks! Released with `10.1.5` :rose:

PR #1019, opened 2016-07-18
Support the "types" compiler option in tsconfig.json.

pwnall:

Definition in commandLineParser.ts:
https://github.com/Microsoft/TypeScript/blob/f19844fbe97656a9e1f743fb6fe4e9d706a83c0d/src/compiler/commandLineParser.ts#L349


basarat:
Thanks! Released with `10.1.5` :rose:

pwnall:
Thank you very much for merging!

PR #1028, opened 2016-07-22
Support tsconfig "typeRoots" compiler option

seanchas116:

This pull request adds `typeRoots` compiler option in tsconfig. http://json.schemastore.org/tsconfig


adamduren:
@basarat @seanchas116 Any update on this? This along with other flags such as `lib` included in `angular-cli` are causing Atom to not be able to load the project.

basarat:
Released this with `10.1.6` :rose:

adamduren:
Thanks!

einfallstoll:
Does this really work? I've defined `typeRoots` as `["./node_modules/@types/"]` and installed `@types/node` but I get an error when using `module.id`. If it's a bug, I'll open an issue.

Issue #848, opened 2016-02-12
TODO Highlight for inline comments


suppayami:
I'm using this pakage for my development, and notice that TODO Highlight only works with block comments, for inline comments like **// TODO: ABC**, it won't work.

This is the screenshot: http://puu.sh/n5hvu/dae2fe49fb.png

sharpmachine:
Same here. Would be nice for it work with `//`

adambuczynski:
See this comment for reference on how to fix this: https://github.com/atom/language-todo/issues/33#issuecomment-228630076

olee:
Would be really great if this could be addressed quickly (as it looks easy to fix according to the link).

googol7:
+1

olee:
I just tested it - replacing [this](https://github.com/TypeStrong/atom-typescript/blob/master/grammars/ts.cson#L1019-L1021) with the following works
````
  "comment-line":
    name: "comment.line.ts"
    begin: "//"
````

olee:
I fixed the previous post with the correct grammer config.
With that it properly recognizes TODOs in TS files.

dsifford:
@olee Did you open a PR with that fix? If not, can you?

olee:
*OT: Why does everyone on open source projects always want reporters to make PRs instead of just integrating the fix themselves as maintainer? :expressionless:*

Sorry, but I'm quite short on time currently so it would be nice if someone else could make a PR (or directly commit to the project if it's a developer).

cortopy:
I don't understand why all the fuss with this issue. Just created a PR with @olee code. It works in my local version

dsifford:
@olee If you're referring to me: I'm not a maintainer.

googol7:
the source of the problem is in TypeScript-TmLanguage and shall be solved there imho. made a pull request there: 

https://github.com/Microsoft/TypeScript-TmLanguage/pull/222

cortopy:
thanks @googol7 looks great

googol7:
The problem should now have been solved with this PR https://github.com/Microsoft/TypeScript-TmLanguage/pull/257 See also this issue: https://github.com/Microsoft/TypeScript-TmLanguage/issues/181

googol7:
Finally this issue has been resolved with release [v10.1.8](https://github.com/TypeStrong/atom-typescript/releases/tag/v10.1.8)

adambuczynski:
Good news, one down, but still many to go :(
[...]

PR #1035, opened 2016-07-26
Feat/language todo support

googol7:
[x] All compiled assets are included (atom packages are git tags and hence the built files need to be a part of the source control)

see https://github.com/TypeStrong/atom-typescript/issues/848

support https://github.com/atom/language-todo single line comments:
    // TODO
    // FIXME
    // CHANGED
    // XXX
    // IDEA
    // HACK
    // NOTE
    // REVIEW
    // NB
    // BUG

PR #1064, opened 2016-09-03
Update ts.cson

cortopy:
[ x] All compiled assets are included (atom packages are git tags and hence the built files need to be a part of the source control)

Change detects TODO-like words so that they can be highlighted. Fixes https://github.com/TypeStrong/atom-typescript/issues/848

### DIFF
--- a/grammars/ts.cson
+++ b/grammars/ts.cson
@@ -1018,7 +1018,8 @@ repository:
     end: "\\*/"
   "comment-line":
     name: "comment.line.ts"
-    match: "(//).*$\\n?"
+    begin: "//"
+    end: '$'
   literal:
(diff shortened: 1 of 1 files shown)


basarat:
This file is generated and should not be hand edited. ref https://github.com/TypeStrong/atom-typescript/blob/master/CONTRIBUTING.md#grammar / https://github.com/TypeStrong/atom-typescript/blob/master/docs/grammar.md Please edit gramar.js  :rose:

basarat:
See https://github.com/TypeStrong/atom-typescript/pull/1035 for another implementation.

cortopy:
Didn't realise this had already been raised. I'll close it

PR #1073, opened 2016-09-22
Update ntypescript to 2016-09-30 release

guncha:

This just updates the ntypescript version to the latest available to get access to some recent Typescript goodies. Due to all Atom plugins running in a restricted environment and and a recent change in Typescript, it's now necessary to `require("typescript")` via a `loophole` (https://github.com/atom/loophole) package provided by the Atom team.

The change causing the issue: https://github.com/Microsoft/TypeScript/pull/9580/files#diff-f620fed640223a55bf6a5c9ceb2e100fR19


guncha:
Due to a recent Typescript PR (https://github.com/Microsoft/TypeScript/pull/11265) the `loophole` hack is no longer necessary! 🎉 

We should still update the Typescript version, though. I'm mainly interested in literal token, number and boolean types.

guncha:
Since Typescript no longer requires the `loophole` hack, I removed it from this PR and updated the `ntypescript` again. That's the only real change. The rest in the generated output are due to changes in how Typescript emits string (double quotes, yay!) and calls to super. Should be 100% backwards compatible.

If it makes sense to you, @basarat, could you please merge?

basarat:
@guncha I've added you to the `atom-typescript-collaborators` team so you should be able to merge / deploy at your will :rose:

guncha:
Thank you, I appreciate the trust. I'll try to look into the other PRs, too.

nycdotnet:
way to go, @guncha !

PR #1079, opened 2016-09-25
Autocomplete does not suggestion options after typing '{'

### LINKED ISSUE #1033: Overly eager autocomplete after typing some characters
Hi,

I've found that the atom-typescript autocomplete is a bit too eager to make suggestions when autocomplete-plus is set to 'Show Suggestions On Keystroke'. As you can see in that attached gif, after typing an opening brace for an 'if' statement the autocomplete pops up and suggests a random symbol to go next. When I press 'enter' to start a new line, instead the autocomplete suggestion is inserted. I have to backspace out the errant symbol. It is a little annoying as it breaks my flow.

animation

I would like to change atom-typescript's autocomplete provider to be a little more conservative about making suggestions when the autocomplete is not explicitly triggered. For instance, it should only make a suggestion if the proceeding fragment could be a variable name. [Or perhaps it would be safer](http://stackoverflow.com/questions/1661197/what-characters-are-valid-for-javascript-variable-names) to just exclude some more characters from autosuggestion!

I had a dig through the code and in the `autoCompleteProvider.ts` file I found this little fragment:

    // line 139
    // if explicitly triggered reset the explicit nature
    if (explicitlyTriggered) {
        explicitlyTriggered = false;
    }
    else { // else in special cases for automatic triggering refuse to provide completions
        const prefix = options.prefix.trim()

        if (prefix === '' || prefix === ';') {
            return Promise.resolve([]);
        }
    }

I would like to extend the exceptions there to include some more characters. Is this a good idea?

fishchisel:

Resolves #1033 'Overly eager autocomplete after typing some characters'. 

I investigated whether it would be possible to only give autocomplete if the user types a character that is valid to start a variable name, but it turns out that's [quite complicated](http://stackoverflow.com/a/9337047/284741), particularly given that Javascript's regex engine does not have any inbuilt definitions of the relevant unicode groups. The quick fix I've added removes 90% of the irritation for me.

### DIFF
--- a/dist/main/atom/autoCompleteProvider.js
+++ b/dist/main/atom/autoCompleteProvider.js
@@ -61,7 +61,7 @@ exports.provider = {
             }
             else {
                 var prefix = options.prefix.trim();
-                if (prefix === '' || prefix === ';') {
+                if (prefix === '' || prefix === ';' || prefix === '{') {
                     return Promise.resolve([]);
                 }
             }
--- a/lib/main/atom/autoCompleteProvider.ts
+++ b/lib/main/atom/autoCompleteProvider.ts
@@ -143,7 +143,7 @@ export var provider: autocompleteplus.Provider = {
             else { // else in special cases for automatic triggering refuse to provide completions
                 const prefix = options.prefix.trim()
 
-                if (prefix === '' || prefix === ';') {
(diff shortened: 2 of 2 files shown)


guncha:
I just realized I can't even type a single opening brace due to Atom automatically entering `{}`, but if it helps you 👍

PR #1091, opened 2016-10-07
Support for tsconfig "importHelpers"

lhecker:

This commit adds support for the new `importHelpers` flag (see [here](https://github.com/Microsoft/TypeScript/pull/9097)).

Interestingly when I ran `npm run build` it changed the 2 grammar files quite heavily, but I didn't include that in this commit/PR.


guncha:
LGTM! Thank you @lhecker 👍

PR #1100, opened 2016-10-14
Fixed usage of deprecated electron env vars

### LINKED ISSUE #1098: atom-typescript automatically opens and focuses to child.js file
The latest atom-typescript opens the child.js from the .atom/atom-typescript/dist/worker and focus to it automatically. There is no such behavior while disabling the plugin.

Atom Version: 1.12.0-beta0
Platform: Windows 10

image

lhecker:

This PR fixes #1098 and adds support for Atom 1.12+. It'd be cool if you guys could release an update ASAP for the other beta testers. 🙂

It was caused by the removal of `ATOM_SHELL_INTERNAL_RUN_AS_NODE` from electron, which in turn has long been renamed to `ELECTRON_RUN_AS_NODE`.

P.S.:
> When I hit this really really annoying bug, it already seemed like something didn't run the `child.js` in a node process like it was supposed to, but instead tried to open or evaluate it which probably caused the file to open with the default file handler (at least that's what I thought). Searching for `child.js` and anything that launches or opens files lead me to `workerLib.ts` and the only `spawn()` in there with it's very suspicious env var. 😂 Best bugfix ever. 10/10 for being much easier than I thought it would be for once.

### DIFF
--- a/dist/main/atom/commands/outputFileCommands.js
+++ b/dist/main/atom/commands/outputFileCommands.js
@@ -40,7 +40,12 @@ function register() {
             else {
                 var command = "node " + path.basename(res.jsFilePath);
                 console.log(command);
-                child_process_1.exec(command, { cwd: path.dirname(res.jsFilePath), env: { ATOM_SHELL_INTERNAL_RUN_AS_NODE: '1' } }, function (err, stdout, stderr) {
+                child_process_1.exec(command, {
+                    cwd: path.dirname(res.jsFilePath),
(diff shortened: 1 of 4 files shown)


guncha:
A fun one indeed! Thanks for debugging @lhecker

Angular-Angel:
So has this change been published yet? I have the atom-typsecript, and it says it's up to date, but I'm still encountering the bug with child.js. Do I need to update manually or something?

Issue #1080, opened 2016-09-26
Typescript highlighting


keithmichelson:
Not sure if this is something for atom or atom-typescript package but when selecting typescript as the file in an Angular 2 project some of the basic javascript doesn't color like you'd think it should but the typescript looks good. If you switch the file type to javascript the regular javascript looks good but not the typescript. Is there a way to get both colored/highlighted?

(select typescript as file type
![screen shot 2016-09-26 at 12 04 01 am](https://cloud.githubusercontent.com/assets/10779020/18824255/a85f22ec-837d-11e6-8ac0-f244d7a47f85.png)

(select javascript as file type)
![screen shot 2016-09-26 at 12 04 18 am](https://cloud.githubusercontent.com/assets/10779020/18824256/abae8906-837d-11e6-85d1-66b453ec7e55.png)

(select typescript as file type)
![screen shot 2016-09-26 at 12 08 35 am](https://cloud.githubusercontent.com/assets/10779020/18824266/b8752b18-837d-11e6-8521-94e2b18bb45d.png)

(select javascript as file type)
![screen shot 2016-09-26 at 12 08 50 am](https://cloud.githubusercontent.com/assets/10779020/18824269/bf5999a0-837d-11e6-844f-6fc72e977418.png)

distante:
I'm starting with typescript and notice this too, is a little hard to keep up with the code when (for example) something like `document.body.innerHTML.length` is just plain white text.

Goodwine:
A **temporary** workaround for now is adding this to your config.cson:

```
core:
  customFileType:
    "sources.js": [
      "ts"
    ]
```

distante:
Should I restart Atom to see the changes?

guncha:
Syntax highlighting got much better in `v10.1.10`. This largely depends on the theme you're using, but at least the proper hooks are there. Give it a shot and let me know how it works for you.

Issue #1061, opened 2016-08-30
Incorrect syntax highlighting after `${xxx}` symbols


yerzhant:
Hi,
maybe the \`${}\` symbols produce this result:
![image](https://cloud.githubusercontent.com/assets/11594067/18073026/abe61cf8-6e84-11e6-9890-d4f412468e3f.png)
i.e.
`
        console.error(msg);
        return Observable.throw(msg);
    }
`
incorrectly highlighted.
Atom    : 1.9.9
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0

guncha:
Fixed in `v10.1.10` 🎉

Issue #1004, opened 2016-07-11
Support syntax highlighting for JSDoc / Typedoc


ddaws:
I would like to document my Typescript code using JSDoc / Typedoc and would appreciate support for syntax highlighting JSDoc comments. I have forked this repo and would like to contribute though would like to understand if anyone is already working on this.

**EDIT :** I am also having trouble building the package via `apm rebuild` because `fs-events` is failing to compile native extensions due to the outdated version of node used by apm. Any advise on this issue would also be appreciated. I have created a [thread here](https://discuss.atom.io/t/upgrading-to-apm-1-11-4-and-encountering-libgit-errors/30901)

guncha:
Basic keyword highlighting was released with `v10.1.10` 🎉 

If you need something fancier, check out https://atom.io/packages/docblockr

PR #1112, opened 2016-10-25
Update grammar files

admosity:
Generated new grammar files to fix/alleviate numerous highlighting issues.

Examples with new grammar from multiple issues:

#1080
image

#1061
image

#1004
image

Upstream changes for decorator highlighting:
Microsoft/TypeScript-TmLanguage#329
image


guncha:
This is great and long overdue! The problem I had last time when trying to do this was that:

1. My Atom theme aggressively colored every identifier red
2. The `.tsx` grammar was broken

Can you please make sure it still works with `.tsx` files?

admosity:
1. It looks great right now with the monokai theme. I think there might need to be some tweaks with some selectors.
2. It doesn't parse at all 😭

admosity:
Looks like the big issue is that the file generated for `.tsx` is truncated. I'll take a look at the script and see what's going on there.

admosity:
image

I loaded the last fuller grammar file defined. This seems to play nicer. I'm not that familiar with `.tsx` syntax, but it seems to be able to parse html tags wherever you place them with this grammar. Looks generally okay.

admosity:
Looking into why that file is so short. Maybe they're combining the two to make the other files so we would have to do the same.

guncha:
It looks like to me that the truncated grammar file only contains JSX related bits and not the whole Typescript grammar, so I think we're expected to reference or include it in the main Typescript grammar to generate a `.tsx` grammar. Perhaps creating a copy of the Typescript grammar, appending all of the grammar expressions from the JSX one and then adding `{ include: "#jsx" }` to the `expression` patterns would do the trick.

admosity:
Yep figured it would be that. I'll look into implementing that later today when I'm not swamped with work.

guncha:
I just had some limited success with it, but it will definitely need some tweaking and automation:

image

Btw, I had to remove the `cast` pattern as it's not valid in a `.tsx` file.

admosity:
Yep I'll load in those xml files and swap out the yaml parser for an xml parser. I'll give that a go later tonight and see the results.

guncha:
@admosity I had some time to do it today and it seems to work ok. I'll open a PR in a sec.

admosity:
Shucks, I'll have to look elsewhere for my hacktoberfest submission. 😆

guncha:
Haha, sorry, this has been on my mind for a while now 😄
Feel free to pick anything from the issue tracker here 😉

PR #1113, opened 2016-10-26
Update grammar script and grammars themselves

guncha:

This PR updates the `scripts/grammar` script to work with the upstream plist files instead of the YAML files as that seems to be the correct way forward according to this:

https://github.com/Microsoft/TypeScript-TmLanguage/issues/333

The script is also now written in Typescript (added ts-node as a dependency) and did some other minor cleanup.

image

The identifiers are still a little too red for my taste, but idk what's going on with that.

Closes #1112


admosity:
What is the theme you're currently using? I think it may be theme related.

guncha:
You are, of course, correct. I'm using `One Dark` and it has a declaration like this:

image

I guess the Typescript grammar was pretty bad before and it didn't tag variables as such, that's why the jarring change.

Thanks for all your help. I think I'll merge this and cut a new release.

chrissm79:
@guncha @admosity This was a HUGE improvement... especially for React (.tsx) files, thank you!!

admosity:
@chrissm79 you're welcome!

PSA: This new grammar does seem to break some existing template string highlighting. So for others that want to have some semblance of highlighting you can add the following to your stylesheet:

```less
  .source.tsx, .source.ts {
    .string.template {
      .variable {
        color: white;
      }
      .punctuation.accessor {
        color: red;
      }
      .template-expression {
        &.begin {
          color: #f92672;
        }
        &.end {
          color: #f92672;
        }
      }
    }
  }
```

So now my highlighting looks like this on atom:
image

PR #1133, opened 2016-11-15
Added new tsconfig options

lhecker:

This adds:
- the es2016, es2017 and esnext targets
- alwaysStrict option
- jsxFactory option

Furthermore:
I wasn't so sure what `typescriptEnumMap` does when a value is not defined:
`"latest"` was only defined in `typescriptEnumMap` but not in the validator. `"es2015"` on the other hand did exist in validator but not in the map.
So I removed `"latest"` entirely and replaced it with the now official `esnext` mode (which is aliased to `latest` - thus backwards compatible).

I also took the liberty to sort the interface and validator keys in c6b7d00 - there are no other changes in that commit.
3f40b96 on the other hand introduces the actual changes above.


guncha:
Thanks!

Yeah, looks like `latest` is not a valid `target` value, while it's used internally by Typescript.

PR #1139, opened 2016-11-22
Update Typescript

guncha:


guncha:
Closing in favor of https://github.com/TypeStrong/atom-typescript/pull/1155

PR #1144, opened 2016-11-24
Improve import quickfix

kyasbal-1994:

I have improved quickfix to import modules.
These are the behaviour I fixed.

## import statement generation for default exported module

In my case, all of my modules are default exported. Therefore, I should remove brackets all time after quick fixed import.
So I append a quickfix to generate not `import {A} from "./A"` but `import A from "./A"`.

I guess there would be other people using default exported modules. 

## path generation for importing js file with `d.ts`

When I use js files with `d.ts`, the path generation for the module is invalid.
Assume there was `A.js` and `A.d.ts` . When I quickfixed that to import , quick fix would generate `import A from "./A.d"`. But it should be `import A from "./A"`

## path generation for external modules not specified as `typings`

If you have module `A`, and package.json of that module specifies `node_modules/A/index.ts` as 'typings',quickfix should generate as `import A from "A"`. I know this feature works fine.

But if there was `node_modules/A/B/C.ts`, that should be able to be refered as `import C from "A/B/C". But, that import would be generated as `../node_modules/A/B/C".


kyasbal-1994:
Is there any problem to merge?

guncha:
Nope. I looked at it yesterday and it seems fine, just need to find some time to test. Will try to get to it this week.

dist/worker/ref/Geometry/Geometry.d.ts
```diff
@@ -0,0 +1,27 @@
+import VertexBufferAttribInfo from "./VertexBufferAttribInfo";
+import AABB from "grimoirejs-math/ref/AABB";
+import Program from "../Resource/Program";
+import IndexBufferInfo from "./IndexBufferInfo";
+import Buffer from "../Resource/Buffer";
+export default class Geometry {
```
guncha: Is this used anywhere?

lib/main/lang/fixmyts/quickFixes/addImportFromStatement.ts
```diff
@@ -9,76 +9,74 @@ import {getPathCompletions} from "../../modules/getPathCompletions";
 
 function getIdentifierAndFileNames(error: ts.Diagnostic, project: Project) {
 
-    var errorText: string = <any>error.messageText;
+  var errorText: string = <any>error.messageText;
```
guncha: Please don't change formatting if you're only changing a few lines in the file. Here, for example, I can't tell for sure if you've added or removed anything.

dist/worker/ref/Geometry/Geometry.d.ts
kyasbal-1994: That is my mistake. Accidentally, my file I using in other project was included in that folder.

guncha:
LGTM, minus the failing specs. Something might have messed up when you were fixing the indentation.

guncha:
👍

Issue #259, opened 2015-04-14
Package option to set import/require completion quote character (single vs. double)


cybrown:
Most of the time, the standard for quotes is simple quotes, is it possible to add an option so the autocompletion can write simple quotes in the file, or maybe set it as a default ?

basarat:
The exact opposite was requested by @pspeter3 : https://github.com/TypeStrong/atom-typescript/issues/203 :smile: 


> is it possible to add an option

I'll change the title of this request to be just that. I'll mark it as up for grabs as I am not motivated by this, I feel either quoting is fine :) PS: this will need some form of support in `snippet`s too.

cybrown:
Ok thank you for your answer :) I understand that it can't be the default behaviour, and that it may be a plugin option.

I think I should be able to do that by myself, but since I do not know how to use the atom API, is the config object a good place to start investigating ? (https://atom.io/docs/api/v0.191.0/Config)

basarat:
> but since I do not know how to use the atom API, is the config object a good place to start investigating

Yes. But we have it neatly abstracted : https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/atom/atomConfig.ts 

Add a new option `preferredQuoteCharacter` with default to `"` and then push it through here : https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/atom/autoCompleteProvider.ts#L217-L235

Coming to think of it its not that hard to do :) I just didn't want to do it as I want all atom projects to look the same. Don't want to be working on one project with one atomts settings and on another with another atomts settings.

But if its what you want (and I am sure there are others) its good to have it :heart: Looking forward to some PR :+1:

basarat:
`atomConfig` is something you can just require anywhere in the project e.g. in `autoCompleteProvider.ts` ;)

park9140:
@basarat, how do you feel about defaulting this based on tslint config?  I assume we should have both tsconfig, and, editor configuration overrides for this.  I was doing some refactoring today and the default being double quote is opposite of our tslint config, I probably spent 20 minutes going back and fixing the incorrect quotes.

basarat:
> how do you feel about defaulting this based on tslint config

That's a great suggestion. I feel very averse to having a package option for it. I've been thinking of putting this in an `ide` section in `tsconfig.json` but using `tslint.json` is *much* better.

cybrown:
[...]

Issue #203, opened 2015-03-23
Import autocomplete uses single quotes


pspeter3:

basarat:
Yes. This is by design. Single quotes for JS double quotes for HTML is what I've learnt to live with based on the teams I've worked with.

pspeter3:
Ok, is this configurable in anyway? Most typescript I've encountered uses double quotes and as result our team has used this convention.

basarat:
@pspeter3 alright the TS compiler uses double quotes : https://github.com/Microsoft/TypeScript/blob/master/src/compiler/scanner.ts plus I trust your judgement. We don't care in our team, so lets just go with double quotes. I'll push.

basarat:
released `v1.6.0` :heart:

pspeter3:
Thanks!

PR #1147, opened 2016-11-24
Add snippet to import using single quotes (as in Angular2)

tiangolo:
Add snippet to import using single quotes (as in Angular2, angular-cli, etc) as `importq`, additional to `import`.

I see this has been debated here https://github.com/TypeStrong/atom-typescript/issues/203 and here https://github.com/TypeStrong/atom-typescript/issues/259.

Right now the checks run OK, but the snippet generates imports with only double quote characters.

Using the `import` snippet generates an import statement with double quote characters (`"`), and if the preferred quote character was `'` it complains that they should be single quote characters. But currently there is no snippet to generate imports with single quote characters from the beginning.

I think this additional snippet might help keep happy both sides (single and double quote "sides").

---

> All compiled assets are included (atom packages are git tags and hence the built files need to be a part of the source control)

Note: there were no changes to TS files.

### DIFF
--- a/snippets/typescript-snippets.cson
+++ b/snippets/typescript-snippets.cson
@@ -19,6 +19,11 @@
         'body': """
             import ${2} from "${1:path}";$3
         """
+    "import using '":
+        'prefix': 'importq'
+        'body': """
+            import ${2} from '${1:path}';$3
+        """
     'importr':
         'prefix': 'importr'
         'body': """


guncha:
Would it be possible to use the *Preferred quote character* from the settings to make sure "both sides" can use the same `import` snippet?

tiangolo:
I wish it was possible, but as far as I know, it isn't.

There doesn't seem to be a way to use variables to customize snippets using the standard Atom snippets.

Here are the references I see relevant: http://flight-manual.atom.io/using-atom/sections/snippets/ and https://github.com/atom/snippets.

---

Nevertheless, I'm one of the "single quoters" (following Angular 2 style) and I would be OK with "having to" use the alternate import snippet (`importq`), actually that's what I'm doing currently with my forked version. And as there is an `import` snippet for double quote imports, and an `importr` for CommonJS imports (`var myModule = require('myModule')`) I wouldn't see it problematic having another "kind" of import snippet.

Also, if there are people using different code styles for different projects (e.g. Angular 2 with single quotes and NodeJS with double quotes) it could be nice to have both snippets available, otherwise it would require them changing the *Preferred quote character* settings each time they switch to work in each project.

guncha:
SGTM 👍

tiangolo:
Thanks for mergin @guncha !

PR #1153, opened 2016-12-07
As per the typescript handbook, allow "none" as a valid module.

### LINKED ISSUE #960: "none" value for "module" compiler option in tsconfig.json considered invalid
Is in fact a valid option and has somewhat different semantics from leaving it unspecified.

kokarn:

Very minimal change but makes the definition follow the handbook.
https://www.typescriptlang.org/docs/handbook/compiler-options.html

Closes #960.


guncha:
Thanks @kokarn 👍

PR #1155, opened 2016-12-07
Update TypeScript to v2.1

timkendrick:

TypeScript 2.1 is now [officially released](https://blogs.msdn.microsoft.com/typescript/2016/12/07/announcing-typescript-2-1/) – this PR updates `atom-typescript` to use v2.1, allowing object spread syntax etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "react": "^0.13.3",
     "season": "^5.1.4",
     "tsconfig": "^2.2.0",
-    "typescript": "2.1.0-dev.20161023",
+    "typescript": "^2.1.4",
     "xtend": "^4.0.0"
   },
   "devDependencies": {


nickfujita:
@guncha I know this was just created a day ago, but would it be possible to get this merged and pushed? I'm excited get a cleanup of all my Object.assigns :)

guncha:
I tried to do this with some nightly version before the `2.1` release and it was broken due to file extension errors that I didn't bother to hunt down. If this works fine, I'll merge it soon 👍

guncha:
`v10.1.13` is out 🎉  As I suspected, there still was an issue with file extensions and I had to finally track it down. If anyone cares, `TypescriptServicesHost` was treating directories as files when it shouldn't have.

I'm looking into porting `atom-typescript` to use `tsserver` which would then take care of keeping track of all the files, parsing and validating `tsconfig.json`, etc. With that kind of setup you'd be able to use whatever typescript is installed in your `node_modules` and this issue (among many, many others) wouldn't be an issue at all. Something to look forward to, I guess.

Issue #985, opened 2016-06-21
Quick fix: import statement for default exports


nielslyngsoe:
Hey guys

I would like to be able to choose an option in the Quick Fix menu that create an import statement but for a default export module.

Currently you can get this:
import {Foo} from 'Foo';

But I would like an option that results in this:
import Foo from 'Foo';

Let me know if you want me to do a Pull Request, but I thought you might have a better way of solving this than my way.

Thanks for the good work

robinisaksson:
+1

nycdotnet:
Please feel free to implement a PR.

PR #1163, opened 2017-01-03
Bugfix/985    quick fix import statement for default exports

andrewicarlson:

---

This PR resolves Issue 985 (https://github.com/TypeStrong/atom-typescript/issues/985) by abstracting a common method used across all of the `addImport...` quickfixes to a utility and creating a new `addImportFromStatementNoBrackets` implementation.


guncha:
We'll be moving to `tsserver` support for quickfixes so this is woefully out of date. Sorry.

Issue #1108, opened 2016-10-19
Deprecated selector in `atom-typescript/styles/atomts-grammar-syntax.less`


Nikki1993:
In `atom-typescript/styles/atomts-grammar-syntax.less`: 

Starting from Atom v1.13.0, the contents of `atom-text-editor` elements are no longer encapsulated within a shadow DOM boundary. This means you should stop using `:host` and `::shadow` pseudo-selectors, and prepend all your syntax selectors with `syntax--`. To prevent breakage with existing style sheets, Atom will automatically upgrade the following selectors:
- `atom-text-editor::shadow .source.ts .require.path,
  atom-text-editor::shadow .source.tsx .require.path,
  atom-text-editor::shadow .source.ts .reference.path,
  atom-text-editor::shadow .source.tsx .reference.path,
  atom-text-editor::shadow .source.ts .es6import.path,
  atom-text-editor::shadow .source.tsx .es6import.path,
  atom-text-editor::shadow .source.ts .amd.path,
  atom-text-editor::shadow .source.tsx .amd.path` => `atom-text-editor .syntax--source.syntax--ts .syntax--require.syntax--path,
  atom-text-editor .syntax--source.syntax--tsx .syntax--require.syntax--path,
  atom-text-editor .syntax--source.syntax--ts .syntax--reference.syntax--path,
  atom-text-editor .syntax--source.syntax--tsx .syntax--reference.syntax--path,
  atom-text-editor .syntax--source.syntax--ts .syntax--es6import.syntax--path,
  atom-text-editor .syntax--source.syntax--tsx .syntax--es6import.syntax--path,
  atom-text-editor .syntax--source.syntax--ts .syntax--amd.syntax--path,
  atom-text-editor .syntax--source.syntax--tsx .syntax--amd.syntax--path`
- `atom-text-editor::shadow .source.ts .keyword.debugger,
  atom-text-editor::shadow .source.tsx .keyword.debugger` => `atom-text-editor .syntax--source.syntax--ts .syntax--keyword.syntax--debugger,
  atom-text-editor .syntax--source.syntax--tsx .syntax--keyword.syntax--debugger`

Automatic translation of selectors will be removed in a few release cycles to minimize startup time. Please, make sure to upgrade the above selectors as soon as possible.

C-Bouthoorn:
This message is also showing on my system:

Atom: `1.13.0-beta1`
[...]

PR #1167, opened 2017-01-12
Bugfix/1108 Update deprecated CSS selector

byrondover:

^ Not applicable for CSS changes, I'm guessing?

Quick fix for Issue #1108.


bbuck:
What about this aspect of the deprecation notice?

> and prepend all your syntax selectors with `syntax--`. To prevent breakage with existing style sheets, Atom will automatically upgrade the following selectors:

petewalker:
Think #1183 supersedes this.

PR #1175, opened 2017-01-16
Remove css shadow selector

lthomaz:
On atom 1.13 the ::shadow css selector was deprecated. 
I just removed it.


dotmagic:
I think just remove may work but isn't the official way to solve it?
`atom-text-editor::shadow ` should become `atom-text-editor.editor`

bbuck:
What about this aspect of the deprecation notice?

> and prepend all your syntax selectors with `syntax--`. To prevent breakage with existing style sheets, Atom will automatically upgrade the following selectors:

PR #1182, opened 2017-02-03
Added styled-components syntax highlighter per bable-language's implementation

### LINKED ISSUE #1181: styled-components syntax support in TypescriptReact
Currently `language-babel` supports syntax highlighting of `styled-components` template strings. Unfortunately, `TypescriptReact` does not. I'd like to request the implementation of this feature 🙂

https://github.com/styled-components/styled-components#atom

JohnLouderback:
* Added styled-components syntax highlighter per `bable-language`'s implementation

This should resolve #1181 

^ No compiled files were changed.

### DIFF
--- a/grammars/css.cson
+++ b/grammars/css.cson
@@ -1,0 +1,1524 @@
+name: "Styled CSS"
+scopeName: "source.css.styled"
+foldingStartMarker: "(/\\*|{|\\()"
+foldingEndMarker: "(\\*/|\\}|\\))"
+fileTypes: []
+patterns: [
+  {
+    include: "#selector-declaration"
+  }
+]
+repository:
+  "attribute-selectors":
+    patterns: [
+      {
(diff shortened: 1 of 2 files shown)


guncha:
The `grammars/ts*.cson` files are generated by the `scripts/grammar` script from the official Typescript language definitions so next time it's rebuilt your changes are going to be erased. The proper solution is to add the rules dynamically somewhere in that script. `css.cson` should probably be in a different package so it can be updated separately and the `ts*.cson` files should contain a single `include: ""source.inside-js.css.styled"` directive. That way if users have the package installed they get the highlighting and if not, they're none the wiser. The scope descriptor should probably just refer to the `language-babel` package.

JohnLouderback:
Will update accordingly

guncha:
Closing due to inactivity. If you or anyone else wants to revisit, I'll gladly review the changes.

PR #1190, opened 2017-02-15
Add functionality to return-from-declaration

deanproxy:


guncha:
Thanks, I looked at the code briefly and it's 👍 . Only thing - I think you also need to add the same command as `typescript:return-from-declaration` to be consistent with how `go-to-declaration` is defined. A keyboard shortcut would also be nice.

If you want, feel free to add this to the beta branch, too!

deanproxy:
Okay, no problem. I'll fix that quickly.

I'll see what I can do on the beta branch as well.

deanproxy:
Okay. I added the new type and a keymap to it. I picked `F10` because it was near `F12` and wasn't already in use as far as my editor was concerned.  symbols-view uses `ctrl-t`

I'll checkout the beta branch tonight and poke around and see what I can do. I'll make a separate pull request for that.

guncha:
Great job, thanks! I'll publish it during the weekend.

PR #1192, opened 2017-02-15
add return-to-declaration functionality on the tsserver branch

deanproxy:

### DIFF
--- a/lib/main/atom/commands/goToDeclaration.ts
+++ b/lib/main/atom/commands/goToDeclaration.ts
@@ -2,6 +2,15 @@ import {commands} from "./registry"
 import {commandForTypeScript, getFilePathPosition} from "../utils"
 import {simpleSelectionView} from "../views/simpleSelectionView"
 
+const prevCursorPositions:any[] = [];
+
+function open(item: {file: string, start: {line: number, offset: number}}) {
+   atom.workspace.open(item.file, {
+     initialLine: item.start.line - 1,
+     initialColumn: item.start.offset - 1
+   })
+}
+
 commands.set("typescript:go-to-declaration", deps => {
   return async e => {
     if (!commandForTypeScript(e)) {
@@ -11,6 +20,8 @@ commands.set("typescript:go-to-declaration", deps => {
     const location = getFilePathPosition()
     const client = await deps.getClient(location.file)
     const result = await client.executeDefinition(location)
+
+    prevCursorPositions.push(location);
 
     if (result.body!.length > 1) {
       simpleSelectionView({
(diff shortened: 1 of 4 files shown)


lib/main/atom/commands/goToDeclaration.ts
```diff
@@ -2,6 +2,15 @@ import {commands} from "./registry"
 import {commandForTypeScript, getFilePathPosition} from "../utils"
 import {simpleSelectionView} from "../views/simpleSelectionView"
 
+const prevCursorPositions:any[] = [];
```
guncha: Missing the type annotation here. Once you add it, you'll discover a type mismatch between what `getFilePathPosition` returns and what `open` expects.

lib/main/atom/commands/goToDeclaration.ts
```diff
@@ -12,6 +21,8 @@ commands.set("typescript:go-to-declaration", deps => {
     const client = await deps.getClient(location.file)
     const result = await client.executeDefinition(location)
 
+    prevCursorPositions.push(location);
```
guncha: It's possible that the user executes this command, gets multiple results, but presses escape and doesn't open any of them. We don't want to record the previous position in that case - only if they make the jump.

lib/main/atom/commands/goToDeclaration.ts
```diff
@@ -27,12 +38,19 @@ commands.set("typescript:go-to-declaration", deps => {
...
+   return async e => {
+      const position = prevCursorPositions.pop();
+      if (!position) {
+         atom.notifications.addInfo('AtomTS: Previous position not found.');
+         return;
+      }
+      open({
+         file: position.file,
+         start: { line: position.line, offset: position.offset }
+      });
```
guncha: The new object you're creating here has the exact same shape as `position` local, so there's no need for that. You could pass it directly, though you'll see that it's not what `open` expects.

deanproxy: Not completely sure what you mean here. position doesn't have a `start` attribute, it looks like:

    {
        file: string,
        line: number,
        offset: number
    }

Open expect the `line` and `offset` to be children of `start`, which is why I wrote it out like that.  Are you saying to just change the definition of `open` to correspond with the `position` argument?

guncha: Oh, gotcha, I didn't see the `start` bit. Carry on 😆

deanproxy:
added those suggested fixes. let me know if you see anything else.

deanproxy:
Just thinking about this... `prevCursorPosition` array can wind up growing too large and consuming memory if you don't return from your go-to's frequently and you're also one of those people like me that keeps their editor open all of the time. This isn't likely to be a big problem in most situations, but something to keep in mind.

guncha:
Thanks @deanproxy 👍

Issue #1191, opened 2017-02-15
Quickfix trigger on line/file


jimmytheneutrino:
Currently quickfixes are only triggered by alt+enter when the cursor is exactly on (or touching) an error. 
This is ok, but sometimes it is much more faster to trigger quick fix when on the same line or even in the same file (usually you only have one or a couple of problems in a file).

What do you think if quickfixes could also be triggered 
1) on a line containing the error,
2) in the same file?

About implementation:

It seems that 1) is a little harder to implement (basically one needs to pass in a query to projectService's getInfoForQuickFixAnalysis not only current cursor position but also current  line's start and ending).

See [here](https://github.com/TypeStrong/atom-typescript/blob/59c4d4dd5ef403b85caea24c7ecaff57045ada95/lib/main/lang/projectService.ts#L819) and [here](https://github.com/TypeStrong/atom-typescript/blob/59c4d4dd5ef403b85caea24c7ecaff57045ada95/lib/main/atom/atomUtils.ts#L11).

Point 2) just needs some adjustments in getInfoForQuickFixAnalysis (or a similar method).

About shortcuts:

I would love if this extended quick fix search could be triggered simply by double alt+enter, but I am not sure if this is doable in atom at all.

jimmytheneutrino:
Hm. I just realized that I could do F8 (to go to next error) and then alt+enter. 

Then might it be best on alt+enter simply to go to the next error first (if there is no error right here) and then show quick fix suggestions?

PR #1193, opened 2017-02-19
quickfix to fall back to closest error in the file

jimmytheneutrino:

See https://github.com/TypeStrong/atom-typescript/issues/1191.
This PR makes quickfix to fall back to the closest error in the current file in case no error is on the position of the cursor.


guncha:
👍

PR #1200, opened 2017-02-22
Add/remove 'typescript-editor' class from <atom-text-editor> elements where typescript is active

### LINKED ISSUE #1199: Add a specific class to the <atom-text-editor> where typescript is active
Adding and removing a specific class (e.g. `typescript-editor`) to the `<atom-text-editor>` elements where typescript is active or not is a good way of keeping track of the availability of certain commands (in keymaps, for example) or it can be used for extra styling customizations.

Using `syntax--*`-like selectors is bad and should be avoided, as, for instance, there could be different packages which use the same syntax grammars but do not provide the same commands.

arthur-xavier:
Fix #1199 by adding the `typescript-editor` class to the editor where `typescript` is activated and removing this class when the editor is destroyed.

### DIFF
--- a/dist/main/atom/editorSetup.js
+++ b/dist/main/atom/editorSetup.js
@@ -1,4 +1,9 @@
 "use strict";
 function setupEditor(editor) {
+    var editorView = atom.views.getView(editor);
+    editorView.classList.add('typescript-editor');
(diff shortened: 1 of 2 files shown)


guncha:
👍

PR #1207, opened 2017-03-06
Quickfix trigger on file: fix off by one error

jimmytheneutrino:

This is a follow up to #1193. See also issue #1191. 

This is a fix for a small leftover bug:
When the cursor is just after the error, passing the current position to the quickfix (e.g., import) won't trigger because the qf's inner filter does not allow that. 

Let's rely on the search for the closest error in this case too.


guncha:
The new version of `atom-typescript` doesn't have quickfixes yet, so this is out of date. We'll make sure it's not a problem with the new implementation.

PR #1238, opened 2017-04-06
Support showing 'unused' ts errors with severity 'info'

### LINKED ISSUE #1018: Warnings vs Errors
Are there any plans to work on emitting warnings next to errors? With the `noUnusedLocals` flag from ts2, such a feature would be interesting.

LaurensRietveld:

I had some issues due to TS and atom-TS treating everything with severity `Error`. 
Particularly situations where you'd like to be notified of the occasional unused variable / parameter, but where you don't want these notifications to clutter your interface that much that you'll miss the more severe TS errors.

Things in this PR are:
- Added an atom-typescript option that shows unused values with severity `info` (defaults to `true`)
- Modified the `errorPusher` file to check the error message, and switch to severity `info` when the above option is set.

Particularly the last item might need some discussion: I'm simply applying a string-match that checks for substring ` is declared but never used`. Ideally I'd like to be able to use TS error codes or something, but afaik the only information I could act on was this error message. This makes it a bit error prone, e.g. when MS changes the error message, or when translations are added to TS or this plugin.
If somebody has a better idea, let me know

ps. this resolves #1018
pps. chose severity `info` instead of `warning` because the default linter colors for `warning` and `error` are too similar in my opinion.

### DIFF
--- a/dist/main/atomts.js
+++ b/dist/main/atomts.js
@@ -39,6 +39,10 @@ function activate(state) {
         });
         subscriptions.add(statusPanel);
         const errorPusher = new errorPusher_1.ErrorPusher();
+        errorPusher.setUnusedAsInfo(atom.config.get('atom-typescript.unusedAsInfo'));
+        subscriptions.add(atom.config.onDidChange('atom-typescript.unusedAsInfo', (val) => {
+            errorPusher.setUnusedAsInfo(val.newValue);
+        }));
         exports.clientResolver.on("pendingRequestsChange", () => {
             const pending = lodash_2.flatten(lodash_2.values(exports.clientResolver.clients).map(cl => cl.pending));
             statusPanel.setPending(pending);
--- a/dist/main/atomts.js
+++ b/dist/main/atomts.js
@@ -39,8 +39,8 @@ function activate(state) {
         });
         subscriptions.add(statusPanel);
         const errorPusher = new errorPusher_1.ErrorPusher();
-        errorPusher.setUnusedAsInfo(atom.config.get('atom-typescript.unusedAsInfo'));
-        subscriptions.add(atom.config.onDidChange('atom-typescript.unusedAsInfo', (val) => {
+        errorPusher.setUnusedAsInfo(atom.config.get("atom-typescript.unusedAsInfo"));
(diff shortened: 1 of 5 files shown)


guncha:
Thanks @LaurensRietveld. I'll have more minor comments on the code later, but to your last point, there's a `code` field on the `Diagnostic` type that should tell you what the specific error is without relying on string matching.

https://github.com/Microsoft/TypeScript/blob/master/lib/protocol.d.ts#L1389

LaurensRietveld:
Ah, completely missed that field. This PR now uses the code (which is the same for an unused local var and an unused parameter)

guncha:
Oh, hello, I guess I missed a notification that you updated the code.

Looks good so far, but I think the `unusedAsInfo` should be passed in by whatever code creates the `ErrorPusher` (`atomts.ts` in this case). This will also allow you to clean up the watcher by adding its dispose function to the existing subscriptions in `atomts.ts`.

`atom.config.onDidChange` also needs a type definition instead of using `(atom.config as any).onDidChange` 😄

LaurensRietveld:
I moved the config subscriber for `unusedAsInfo` to `atomts.ts`. 

About the type def:  I could add it to `atom.d.ts`, but we'd need to overwrite the type defs of the already declared var `atom` (done in the `@types` atom ts def) with an extended `IApp` and `IConfig` interface. Not sure whether this is easily/cleanly doable

guncha:
You can only extend types in Typescript, never overwrite. For example:

```ts
interface Foo {
  bar: Bar
}

var foo: Foo

// elsewhere, but in the same scope..

interface Foo {
  baz: Baz
}
```

The type of the `foo` variable will have both `bar` and `baz` props. You have to do the same thing here, find where `IConfig` is defined and add the necessary types in either `atom.d.ts` or perhaps `atom_core.d.ts`.

guncha:
Oh, also please run your code through the `format` command and remove the semicolons to match the style of the rest of the source 😄

LaurensRietveld:
I tweaked the style and added the interface

guncha:
Looks good 👍  I'll test it and merge today or this weekend.

guncha:
Nice, thanks for contributing, @LaurensRietveld! 🎉

Issue #1161, opened 2016-12-19
Reference path autocompletion not working anymore.


rzvc:
Since the last update, /// &lt;reference path="" > stopped using autocompletion for paths.

rathermarv:
Have also found this. Version 10.1.9 appears to be last stable version for this functionality.

rzvc:
It looks like this functionality was removed when the autocompletion provider was rewritten.

The old one seems to have had a special case for file autocompletion, which was handled by getting all the relevant, relative paths using getRelativePathsInProject() (which is still there, but not used).

The current one is relying on tsserver to provide the autocompletion suggestions, but tsserver doesn't do that.

I wanted to add it back, but it seems that projectService.js (the file that contains getRelativePathsInProject() is heavily relying on some "transformer" file, which doesn't seem to be around anymore and neither the functions it used to expose. This is where I gave up.

If anyone wants to pick this up, please note that the importPathScopes is wrong, it says 'tripple-slash-directive' and it has to be 'triple-slash.directive' in new versions of Atom. Also, `this.disableForSelector = ".comment";` should be commented out, because the triple slash directives also gain the .comment class and that would make the autocomplete provider not work on them.

I hope this gets bumped in priority because, without it, writing relative paths by checking the tree structure is extremely difficult.

guncha:
Thanks for looking into it, @rzvc. You're right, this is one of the things I removed during the switch to `tsserver` based completions. `getRelativePathsInProject` (sorry about the dead code, btw) relied on the internals of the compiler to get the file list which we don't have access to anymore. I also don't use triple slash directives so this was never a priority. The regular module completions work via `tsserver` though there is much to be desired.

If you or anyone is interested in implementing autocomplete for triple slash references, there's a `projectInfo` command that can return a list of all files in the project. That plus changes to the `disableForSelector` and detecting the type of completion the user is asking for in `getCompletions` would get you mostly there.

hszhsh:
@rzvc @guncha I've added this functionality back using projectInfo.body.fileNames, and committed here: https://github.com/hszhsh/atom-typescript/commit/d82bd1dd74ee43c7a9dd7fa214dc8392a0eba2b3. Please have a review if this approach is alright. I'd like to reorganize the code and contribute to the project.

guncha:
Thanks @hszhsh! Please open a PR and I'll review it in the next few days.

PR #1246, opened 2017-04-19
add refrence path auto completion

hszhsh:
https://github.com/TypeStrong/atom-typescript/issues/1161

Please have a review if this approach is alright. I'd like to reorganize the code and improve this functionality further.


lib/main/atom/autoCompleteProvider.ts
```diff
@@ -88,6 +89,53 @@ export class AutocompleteProvider implements Provider {
...
+
+  async getPathSuggestions(
+    prefix: string,
+    location: FileLocationQuery): Promise<SuggestionWithDetails[]> {
+    const client = await this.clientResolver.get(location.file)
+    const projectInfo = await client.executeProjectInfo({
+      file: location.file,
+      needFileNameList: true
+    })
+    const filePaths = projectInfo.body!.fileNames!.filter(p => p != location.file);
```
guncha: The optional body for the `project-info` request is there for a reason - Typescript server might not always know what project the current file is in. It's up to you to handle it gracefully.

lib/main/atom/autoCompleteProvider.ts
```diff
@@ -88,6 +89,53 @@ export class AutocompleteProvider implements Provider {
...
+    var files: {
+      name: string;
+      relativePath: string;
+      fullPath: string;
+    }[] = []
+    var sourceDir = path.dirname(location.file);
+    filePaths.forEach(p => {
+      let relativePath = path.relative(sourceDir, p).split('\\').join('/');
```
guncha: Is this meant to convert from windows paths to unix? If so, there are specialized functions in `path` module for this.

lib/main/atom/autoCompleteProvider.ts
```diff
@@ -88,6 +89,53 @@ export class AutocompleteProvider implements Provider {
...
+    return files.map(file => {
+      let suggestionText = file.relativePath;
+      let suggestion: Suggestion = {
+        text: suggestionText,
+        replacementPrefix: prefix,
+        rightLabelHTML: '<span>' + file.name + '</span>',
+        type: 'import'
+      }
+
+      return suggestion;
```
guncha: This whole function is messy. You're mixing `const` and `var` and mapping over arrays multiple times and creating unnecessary intermediate objects. You should also separate the relative path logic from generating and filtering the suggestions.

guncha:
Also, sorry about taking so long to look at this. I'll try to be better! 😄

hszhsh:
Thanks for you review. I copied most of the code from the old project with little modification, so it's really sprawling. 
For the import path issue, actually lib/util is what I need instead of ../../../../../../lib/util. Because I'll merge multiple paths when my project deployed, and I'd like absolute paths from the path roots instead of relative paths. Is there any other way I can configure it to behave like this?

hszhsh:
Thanks for you review. I copied most of the code from the old project with little modification, so it's really sprawling.
Yes, paths in tsconfig is not handled properly. I think I'll open another project and try to fix this. And I'll try to refactor the code.

guncha:
I think for the time being we can make sure this works for reference comments only and use the default completion for module imports.

P.S. I just reformatted the whole codebase using `prettier` so you're getting some conflicts. I recommend cherry-picking the commit that adds prettier, running it on the codebase and only then merging the master.

hszhsh:
OK, I'll limit this to reference comments and think about module imports further.

lib/main/atom/autoCompleteProvider.ts
hszhsh: I cannot find such functions in path module, but found makeRelativePath function in main/atom/utils/fs module which does the same thing in the same way.

hszhsh:
Hi, I noticed that you specified ".comment" as disableForSelector for AutocompleteProvider, so it won't work in reference comments. Is it ok to remove the disableForSelector or I should place the code somewhere else?

hszhsh:
And I also find that the code cannot handle /// <reference types="..." /> properly. The issue is the scopes are the same with scopes of /// <reference path="..." />.

guncha:
The `disableForSelector` is just there to cut down on the calls to the `getSuggestions` function. You can safely remove it as long as you make sure the function returns early if it's not a triple slash reference comment.

Issue #1234, opened 2017-04-03
Hyperclick support


andreasonny83:
Not sure why but the `cmd+click` shortcut used to work for me on Atom time ago with [atom-hyperclick](https://github.com/facebooknuclide/hyperclick) installed but now doesn't seem to work any more.
It would be great if I can jump to my Tytpescript declarations just with `cmd+click` instead using `fn+F12` from keyboard.

Thanks

olee:
This happened to me as well, after updating to the new atom-typescript v11.
If you go back to the latest v10 release, it works again.

I guess I will wait with updating, until this and quickfix work again.

andreasonny83:
It's working again with the latest Atom version (1.16)

sverrejoh:
@andreasonny83 Does it work for you still? Did you have do do something special to get it to work? On my machine it does not work.

guncha:
This should still be an open issue since there's no hyperclick support in the newest version.

andreasonny83:
@sverrejoh, I notice that using `Cmd + F12` with the `atom-typescript` plugin works better than `Cmd + click` with `Hyperclick`. However, in order to use the hyperclick feature, I think you still need to install the `goto` and `goto-definition` plugins on top of `Hyperclick`

guncha:
This is probably a question of taste. I prefer Cmd + F12 too, but other people have told me that they prefer to use mouse for this. Since the hyperclick provider barely adds any overhead, I think we can easily have both.

sverrejoh:
I agree keyboard shortcut is usually nicer (I remap it as I don't have a  F12 button on my keyboard), but I'm used to cmd+click from using FlowType, and I kind of like it too.

I see! it works together with goto-definition. I'll try that if I'm not able to get the old build in code to work again. Thank you!

sompylasar:
'Go to declaration' works for me in `atom-typescript@11.0.4` via `hyperclick` configured for `option+click` (non-default) on macOS. Just checked, `cmd+click` can also be used if configured in `hyperclick` settings. Looks like https://github.com/TypeStrong/atom-typescript/pull/1265 is a success here. https://github.com/TypeStrong/atom-typescript/compare/v11.0.3...v11.0.4

pingshunhuangalex:
It stops working on Atom 1.21.0-beta1. Has this change been implemented yet?

bhperry:
[...]

Issue #1253, opened 2017-05-10
Assign "Go To Declaration" to Ctrl + Click


josh18:
Is it possible to make the "Go To Declaration" method fire when your Ctrl + Click on a reference?

If not, is it possible to add this as a feature?

sverrejoh:
Hyperclick should provice this, and atom-typescript supposedly has support, but I'm not able to get it to work. #1234 is marked as resolved, but it seems like it might have broken.

guncha:
That's right, we had support for it, but it got lost in the shuffle when switching to the new compiler API. It should be pretty easy to add it back again - it just needs some plumbing to connect the hyperclick provider with the `go-t-declaration` command. If someone wants to take a stab at it, I'll gladly review the PR.

sverrejoh:
@guncha I created a PR for it, doing the same as in `go-to-declaration`. Not sure if there's a nice way to nicely unify the code in the two files, but this works at least.

sverrejoh:
I fixed some issues and updates the pull request. 

https://github.com/TypeStrong/atom-typescript/pull/1265

sompylasar:
'Go to declaration' works for me in `atom-typescript@11.0.4` via `hyperclick` configured for `option+click` (non-default) on macOS. Just checked, `ctrl+click` can also be used if configured in `hyperclick` settings. Looks like https://github.com/TypeStrong/atom-typescript/pull/1265 is a success here. https://github.com/TypeStrong/atom-typescript/compare/v11.0.3...v11.0.4

PR #1265, opened 2017-05-23
Added back support for Hyperclick

### LINKED ISSUE #448: Implement hyperclick package

sverrejoh:
Hyperclick support dissapeared after a refactor.

Added the old code back, and updated it to work with the new API, as
used in `go-to-definition`. Included support for selecting between
multiple definitions.

Should fix #448 and #1253


splodingsocks:
Awesome! I was really missing hyperclick. One issue I just found, though. If you use hyperclick to navigate to a symbol definition, `F10` doesn't return from the definition like it does when you use `F12` to go do the definition. It shows this error instead:

atom beta_2017-05-31 13-56-33

sverrejoh:
Right, I saw some state related to that in the jump-to-definition code. This should probably be unified somehow, and I can see what I can do :). Thanks for the feedback!

sverrejoh:
@splodingsocks Thank you again for the feedback! 

I fixed the issue with returning to definition with `F10` and updated the pull request. Should be more solid now.

splodingsocks:
Nice!!

sverrejoh:
@guncha Am I missing something in this pull request? I'm happy to clean it up if anything's missing.

jonyeezs:
this should fix #1234 as well

lib/main/atom/hyperclickProvider.ts
```diff
@@ -0,0 +1,30 @@
...
+import {simpleSelectionView} from "./views/simpleSelectionView"
+import {handleDefinitionResult} from "./commands/goToDeclaration"
+
+export function getHyperclickProvider(clientResolver: ClientResolver): any {
+    return {
+        providerName: "typescript-hyperclick-provider",
+        wordRegExp: /([A-Za-z0-9_])+|['"`](\\.|[^'"`\\\\])*['"`]/g,
+        getSuggestionForWord(editor: AtomCore.IEditor, text: string, range: TextBuffer.IRange) {
+            const TS_GRAMMARS = ["source.ts", "source.tsx"];
+            if (TS_GRAMMARS.indexOf(editor.getGrammar().scopeName) === -1) {
```
guncha: This should use `isTypescriptGrammar` helper from `typescriptEditorPane.ts`.

lib/main/atom/commands/goToDeclaration.ts
```diff
@@ -56,3 +36,26 @@ commands.set("typescript:return-from-declaration", deps => {
       });
    }
 });
+
+export function handleDefinitionResult(result: protocol.DefinitionResponse,
+                                       location: FileLocationQuery): void {
+  if (result.body!.length > 1) {
```
sverrejoh: I agree. I just copied it, but then it was in a promise, so I guess we can argue the exception was handled :).

Fixed it now.

lib/main/atom/hyperclickProvider.ts
sverrejoh: Great! I extracted it and put it in utils.

guncha:
Solid PR! 👍 

I could have sworn I wrote a comment about using the scopes from the Typescript grammar to narrow down the locations where Hyperclick is effective. Basically, when you return from the function, the `range` property is used to highlight with an underline the section of code to let the user know "hey, if you click here, I'll take you somewhere". Currently, it will highlight every word, including string literals, and not everything, when clicked, will result in a jump to definition.

That's something that would take this from 95% to 100%, but it can be a part of a separate PR. I'll release this so people can start using it.

sverrejoh:
Great suggestion 😊

I started looking into the scopes, and have something basic running. I'll try to come up with a list of what's not hyperclickable, and I can push it as a separate pull request. I think I need to add `scopeDescriptorForBufferPosition` to the TypeScript atom definition over at DefinitlyTyped for this as well.

From the top of my head:
- String literals outside import statement
- all other primitive types
- keywords
- within comments

Thanks a lot!

guncha:
Sounds about right. You can always add it to the `typings/atom_core.d.ts` and then, if you want to be nice, open a PR against the DefinitelyTyped definitions.

Issue #1279, opened 2017-06-20
Semantic View integration is missing


russaa:
similar to issue #1219 "Symbols-view integration is missing", the _Semantic View_ (or _Outline_) is also missing since the update to v11.x.x

cf. the image from the old (v 10.x.x) [README](https://github.com/TypeStrong/atom-typescript/tree/v10.1.15#semantic-view) (Semantic View is on the right):
![](https://raw.githubusercontent.com/TypeStrong/atom-typescript-examples/master/screens/semanticView.png)

I find this feature extremely useful -- in fact, its absence is _the_ reason for me, why I did not switch to v11.x.x yet (closely followed by the missing symbols-view)

lierdakil:
Released as v12.2.0 :tada:

PR #1280, opened 2017-06-20
Re-Adding Semantic View #1279

russaa:

This modification re-adds the [Semantic View](https://github.com/TypeStrong/atom-typescript/tree/v10.1.15#semantic-view), incl. the `Toggle Semantic View` command -- see issue #1279.

Basically, I re-used the semantic view from v10.x.x, creating the "AST" information for the current file via the typescript `"navtree"` command.

I am new to atom plugins/packages, so I am not sure, what framework should be used for creating views -- there was some comment, that React will be "phased out", so I replaced that part with _pure_ HTML.
I am not really happy about this, e.g. its performance could be improved using a framework that optimizes HTML-access & reuse of HTML-elements.

Also I used the `TypescriptBuffer` instead of "manually" opening a file via the typescript server commands -- I am not sure if that's the way to go.


lib/client/client.ts
```diff
@@ -103,6 +104,9 @@ export class TypescriptServiceClient {
   executeSaveTo(args: protocol.SavetoRequestArgs) {
     return this.execute("saveto", args)
   }
+  executeNavTree(args: protocol.FileRequestArgs) {
```
guncha: You should specify the return type here. It's something like `Promise<protocol.NavTreeResponse>`.

lib/main/atom/atomConfig.ts
```diff
@@ -0,0 +1,31 @@
...
+function getConfig<T>(nameLambda: () => any): T {
+    return atom.config.get(packageName + '.' + getName(nameLambda));
+}
+function setConfig<T>(nameLambda: () => any, value: T): T {
+    return (atom.config as IConfig).set(packageName + '.' + getName(nameLambda), value);
+}
+
+class Config {
+    schema = {
+        showSemanticView: {
```
guncha: The semantic view probably doesn't need a setting if we can make it non-intrusive and easy to toggle when necessary.

lib/main/atom/views/semanticView.tsx
```diff
@@ -0,0 +1,424 @@
...
+        this.selectAtCursorLine();
+      });
+
+
+      editorChanging = editor.onDidStopChanging(() => {
+        //set navTree
+        const filePath = editor.getPath();
+        this.getTypescriptBuffer(filePath).then(buffer => {
+          if (buffer.isOpen) {
+            this.loadNavTree(buffer, filePath);
```
guncha: To answer your question, you only need to use the TypescriptBuffer wrapper if you plan to make changes to the source file. This is so `tsserver` is kept in the loop of any changes and can update its internal representation. In this case you don't even need to know the content of the file so it's not necessary.

lib/main/atom/views/semanticView.tsx
```diff
@@ -0,0 +1,424 @@
...
+ */
+namespace rts {
+  /** 0 based length */
+  export function indent(indent: number) {
+    return Array(indent + 1).join().split('').map(i => "\u00a0\u00a0\u00a0\u00a0");
+  }
+}
+
+//TODO extract?
+interface Position {
```
guncha: Already in `utils` under `LocationQuery`.

lib/main/atom/views/semanticView.tsx
```diff
@@ -0,0 +1,424 @@
...
+import * as dom from "jsx-render-dom"
+import {TypescriptBuffer} from '../../typescriptBuffer';
+
+/**
+ * Eventually
+ */
+namespace rts {
+  /** 0 based length */
+  export function indent(indent: number) {
+    return Array(indent + 1).join().split('').map(i => "\u00a0\u00a0\u00a0\u00a0");
```
guncha: This is silly. Indentation should be done with CSS.

lib/main/atom/views/semanticView.tsx
```diff
@@ -0,0 +1,424 @@
...
+  }
+}
+
+//TODO extract?
+interface Position {
+  line: number;
+  offset: number;
+}
+
+interface SemanticTreeNode {
```
guncha: Also defined in `typescript/lib/protocl` as `NavigationTree`.

lib/main/atom/views/semanticView.tsx
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  forceUpdate() {
+    //FIXME should this be done differently?
+    this._render();
+  }
+
+  loadNavTree(buffer: TypescriptFileBuffer, filePath?: string) {
+    filePath = filePath ? filePath : this.editor.getPath();
+    buffer.buffer.getClient(filePath).then(client => client.executeNavTree({file: filePath as string}).then(navtreeResult => {
```
guncha: Like I mentioned elsewhere, you just need to get the client here which is something that `ClientResolver` can do for you. Basically you have to replace `getTypescriptBuffer` with an instance of `ClientResolver` provided by `atomts.ts`.

russaa:
I think/hope, I made all the detail-changes as requested.

but w.r.t. issue 1. _(semantic view as a "normal" Atom panel/Dock)_:

it seems to me, that other "outline views" (e.g. symbols-tree-view) use a similar approach as is currently used, i.e. do not insert the panel into a dock.

I think the reason may be, that this way, the panel can be easily toggled/hidden, when no appropriate content/file is opened -- i.e. each outline-view shows its outline, when there is parse-able content, and hides itself, when there is not.

With the dock-approach, this could (or better should) not be done, since hiding the dock is inappropriate, as there may be other panes/items in the dock.  
Also "auto-focusing" the semantic-view panel in the dock (e.g. in case a typescript file is focused in the editor, and the dock is open, but another pane/item is currently shown) would probably be not a good idea, since users may get annoyed, if it changes automatically, although they want to use a different view/panel/item in the dock -- maybe if such behavior were configurable, it would be acceptable(?)

So I think, for this to work nicely (in concert with other outline views), there should be some kind of general outline-view-interface, where each outline-view would register, and then show its outline, if appropriate content is shown in the focused editor. Otherwise users may be stuck, constantly switching outline-views in the dock, when they are working on different file-types.  
Unfortunately, I am not aware of any such interface, or proposal for such an interface ...

So I think for now, the current approach may actually be the "more practical" solution, at least for the time being.  
What do you think?

btw. I did start integrating the panel into a dock, but ran into the mentioned _usability_ problems -- I could upload the changes into a different branch, if you want to have a look/try yourself

russaa:
hmm, `prettier` cannot be run out-of-the-box on Windows systems, but even after fixing this, i.e. getting the prettier-script to run _(just a note: commandline-arguments instead of env-variables would probably be more system-independent)_:  

`npm run test` (run on my system) says now
```
Checking for files that need formatting:
All good.
```

while the travis-ci still says:  
```
Checking for files that need formatting:
./lib/client/client.ts
./lib/main/atom/commands/semanticView.ts
./lib/main/atom/views/semanticView.tsx
```

I am not sure what to  do, since I did run prettier on my system (several times) and on testing, it says everything is OK ... how can I find out, what exactly needs changing?

guncha:
What's the reason for not running on Windows as is?

It might be a different version of `prettier` or the settings given to it might be different. 

One thing you can do is replace the test script in the `package.json` with `npm run prettier; git diff` to run prettier and then show you any changes that it made.

russaa:
in Windows, the environment variables are set differently (with keyword `set`), so the command-lines needed to be changed to
```
"prettier": "set ACTION=write&& ts-node scripts/prettier.ts",
"test": "tsc --noEmit -p ./lib && tsc --noEmit -p ./scripts && set ACTION=check&&ts-node scripts/prettier.ts"
```

Anyway, `git` does not report any differences. Which is not surprising, since the test script claims that everything is formatted OK
```
...
> tsc --noEmit -p ./lib && tsc --noEmit -p ./scripts && set ACTION=check&&ts-node scripts/prettier.ts

Checking for files that need formatting:
All good.
```

I will try to set it up on a *nix machine, maybe that will help.

russaa:
OK, `prettier` on *nix did work -- and I think I found the problem with Windows:  
in the prettier script, the last argument (file-glob, [line 42](https://github.com/TypeStrong/atom-typescript/blob/master/scripts/prettier.ts#L42)) needs to be enclosed in double-quotes, instead of single quotes (single quotes only work on *nix, double-quotes work on *nix and Windows), i.e. instead of
```javascript
...
"'./{lib,scripts}/**/*.{ts,tsx}'"
...
```
the line needs to be
```javascript
...
'"./{lib,scripts}/**/*.{ts,tsx}"'
...
```

-----

__But back to issue at hand:__  
do you have an opinion yet, about semantic-view in a Dock vs. its current integration?

guncha:
Let me see if I understand the alternatives here:

1) The symbol view is attached to each Typescript pane and not shared between other panes. 
2) The symbol view is a shared panel that shows the symbols from the active pane only (or nothing since the user might be working on a non-Typescript file)

With the earlier comment I made, I was mainly concerned about having the ability to resize the symbol view and close it with the mouse. With that in place, the per-pane symbol view could be a more flexible solution.

guncha:
Btw, I'm looking into using something like Preact for the status bar to make the code more declarative. If that works ok, we should rewrite the semantic view code to use it too. What do you think?

russaa:
I am not really a React expert, but from what I can see an their github page, Preact looks like a promising lightweight alternative -- if it's good enough for you, then I would "migrate" the semantic-view to that solution.

russaa:
w.r.t. your questions:
> 1. The symbol view is attached to each Typescript pane and not shared between other panes

not exactly: the symbol-view would be inserted as "right panel" to the workspace, i.e. it's on the very right side of the workspace, but it's only one instance/pane.  
So in this sense, actually both alternatives would be a "shared panel" for the Typescript panes.

> 2. The symbol view is a shared panel that shows the symbols from the active pane only (or nothing since the user might be working on a non-Typescript file)

Yes, where the main difference (when there's no appropriate content in the focused editor) to alternative 1 would be, that the dock/semantic-view would stay open/visible.  
I.e. it would show nothing/be useless at that moment for the user.

But thinking about it, the dock-variant (i.e. alternative 2) would definitely be more clean, and more in line "as it should be done":  
e.g. the "right panel" of the workspace is actually left to the right dock. So if using alternative 1, and the user has the right-side dock open, the semantic-view would appear next to the dock's left side, and would "steal" additional space (in addition to the right dock), e.g.  

... user working on non-Typescript content
```
|                 |            |
| workpace/editor | right dock |
| (non-Typescript)|            |
|                 |            |
```
... and when the user opens Typescript content (alternative 1):
```
|                 |                 |            |
| workpace/editor |   right panel   | right dock |
|  (Typescript)   | (semantic-view) |            |
|                 |                 |            |
```

... I guess this means, that we should go with alternative 2, right?

guncha:
Yeah, I think we should go with the dock variant because it allows users to move the panes around and customize their workspace however they see fit. We also get resizing and close using the mouse for free. Basically, let's copy what the `tree-view` and and the git tab packages are doing.

russaa:
The semantic view now loosely follows the design of `tree-view`:  
e.g. the created HTML (and thus the styling) is roughly the same as `tree-view`, minus some fancy stuff like listening to theme-changes, drag&drop/moving or renaming -- that is not supported (yet?).

There is an experimental/_hacky_ feature implemented for toggling the collapse/expand state in the semantic tree view:  it use explicit knowledge of the expand/collapse-icon, so that it can toggle by clicking the icon (clicking the entry itself will jump/navigate to its position in the text editor).  
The reason for this is, that in the `tree-view`, expand/collapse is toggled by clicking the whole item, which would not be good solution for the semantic-view, since in the `tree-view`, toggling is only active for directories, and files are only opened (i.e. cannot be expanded/collapsed).  
But in the semantic view, the entries that can be expanded/collapsed can also be "opened" (that is, jumping to their position in the text editor).  
So, there needs to another mechanism for toggling the entries in the semantic view with the mouse (another than clicking the whole entry).  
Unfortunately, the expand/collapse icons are "inserted" via pseudo elements `::before`, so they cannot be accessed directly via javascript -- hence the HACK, using the pixel-size for detecting, if the expand/collapse icon was clicked, or if the entry itself was clicked.  
_I don't know if there is a better way for accomplishing the toggling, without abandoning use of the `tree-view` layout/CSS -- if you have any ideas, I would be grateful for the input._


One other major difference would be, that `tree-view` has a "heavy" data-model, that is, a data-model which has its own change-detection.

I.m.o. it would not make much sense, to realize something like that for the semantic view, since `tree-view` can leverage the file-system etc. for detecting changes.  
So I think, the better/less involved strategy for the semantic view would be, to use a change-aware rendering engine (like React/Preact).

But for the time being, the semantic view's HTML is simply recreated on every change.

The next steps, that I see are: 

 1. deciding, if the simple HTML-creation should be replaced by a rendering engine, e.g. `Preact` that you suggested.  
     Is there some news yet, if we should use `Preact`?
 2. currently, all classes are gobbled up in `semanticView.tsx`; they could be refactored into (probably 3) separate files.  
    What do you think, should that be done?

lib/client/client.ts
russaa: done

lib/main/atom/atomConfig.ts
russaa: first I did remove the config, but now is re-added (in `lib/main/atomts.ts`):  
I see no other way, how we could remember the user's wish/setting, when Atom is started the next time.

I.e. the plugin should remember, if the user did have the Semantic View open (or not), the next time, Atom is started.

lib/main/atom/views/semanticView.tsx
russaa: done (removed / replaced by `tree-view`'s CSS styling)

lib/main/atom/views/semanticView.tsx
russaa: done (changed to use the existing type)

lib/main/atom/views/semanticView.tsx
russaa: done (changed to use the existing type)

lib/main/atom/views/semanticView.tsx
russaa: done (replaced use of `TypescriptFileBuffer` by `ClientResolver`)

lib/main/atom/views/semanticView.tsx
russaa: changed (use `ClientResolver` now)

russaa:
@guncha any news?

I.e. on  how to proceed for re-adding the semantic-view?

... in the mean time also added a symbols-view for searching the file-symbols:  
basically it is a copy of the relevant code from `symbols-view`, but instead of using `ctag`, it uses the typescript navtree command for populating the symbols list.

I did not commit these changes yet though, since I  did not know, if that should be added here, too.

lierdakil:
FWIW, the preferred UI framework for Atom seems to be https://github.com/atom/etch -- which is similar to React in design, but is somewhat lighter. I've been using it with https://github.com/atom-haskell/ projects, and it's rather nice. More flexible than React in some aspects even, and is somewhat leaner.

lierdakil:
Okay, so as a heads-up, I just merged PR #1393, which drops space-pen and jsx-render-dom in favour of etch everywhere. It's what Atom uses internally, so we might as well (besides, it's rather nice). If you're still interested in this PR, I suggest rebasing on `master` and converting `semanticView.tsx` to use etch -- also probably a good idea to split that behemoth into separate components. I'll do my best to help merge this. Terribly sorry it got neglected.

russaa:
thanks so much for the update

it will probably take me a week or two, but I will look into updating this PR & then will post any updates I have

lierdakil:
Note: CI is probably failing because you need to update typescript on your machine, it basically complains that js files *it* built don't match what's committed.

russaa:
... yes I did not commit the compiled files -- now they are included.

The code is not pretty (and probably includes some unnecessary/unused code), but it should work

russaa:
Sorry to spam this thread, but I have a question about the `lib/tsconfig.json` file:  
would it be possible to set `noUnusedParameters` to false?

Otherwise the compile will complain about unused parameters (in public functions) -- since some of these or API/interface functions.
If this warning stays enabled, there would need to be some dummy-code that "touches" the parameters, otherwise it won't compile (removing the parameters is not an options, since the functions implement/specify an interface)

lierdakil:
`noUnusedParameters` can be suppressed by appending `_` to parameter names.

lib/main/atom/views/semanticViewPane.ts
```diff
@@ -0,0 +1,120 @@
...
+        atom.workspace.getPaneItems().forEach(paneItem => {
+          if (paneItem instanceof SemanticView) {
+            paneItem.destroy()
+          }
+        })
+      },
+    })
+
+    this.subscriptions.add(
+      atom.workspace.onDidAddPaneItem((event: PaneItemObservedEvent) => {
```
lierdakil: This definitely looks weird. If I closed/opened the semantic view, it doesn't necessarily mean the settings should change, does it? Besides, tracking the events doesn't make all that much sense either, because we're fully in control of when those events happen (via `show`/`hide`/`toggle`).

Same with `onDidDestroyPaneItem`

lib/main/atom/views/semanticViewComponent.tsx
```diff
@@ -0,0 +1,544 @@
...
+  }
+
+  private getNodeEndLine(node: NavigationTree): number {
+    const s = node!.spans
+    return s ? s[s.length - 1].end.line - 1 : 0
+  }
+
+  private getDomNodeStartLine(elem: HTMLElement): number {
+    // return parseInt(elem.dataset.start as string, 10)
+    return parseInt((elem as any)["data-start"] as string, 10)
```
lierdakil: Why not the first option though?

lierdakil:
I took the liberty of going over the code once and cleaning it up a little bit, I hope you don't mind.

I still don't fully comprehend the purpose of `onDidAddPaneItem` and `onDidDestroyPaneItem`, those look out of place. Seems like a poor man's serialization to be honest.

General considerations:

* You probably could split actual nodes into a separate Etch component. That would likely make the code a lot easier to read.
* Nodes seem to be a dangerous mix of representative and model layers, hence all the DOM traversal in `findNodeAtCursorLine`. Wouldn't it be easier to keep model and representation separate, walk over the model and then `etch.update` to highlight the corresponding component? It's usually a bad idea to intermix direct DOM manipulations with either React or Etch since that diminishes the benefit of using those frameworks a lot.

russaa:
thanks for the review -- I'll revise the code and resubmit.

Since I haven't done much with Atom yet --and etch/react for that matter-- every helpful tip is welcome.

lib/main/atom/views/semanticViewPane.ts
russaa: I guess it depends on how you want the view to behave:   
I'd prefer that my last decision would be remembered, when restarting Atom, instead of having to open the settings and change it there.

But that is probably a matter of taste -- there should probably be an additional setting for controlling that, i.e. if the editor should remember my last decision, or use a fixed setting for opening/not opening the semantic view the next time Atom is started(?)

But I definitely move it into the show/hide (or better yet serialization/deserialization hooks) -- I wasn't sure, if Atom has some other way of controlling the view, so the listeners seemed the safest way to catch opening/closing of the view.

lierdakil: > I'd prefer that my last decision would be remembered, when restarting Atom, instead of having to open the settings and change it there.

It's not necessarily a decision. I.e. you can't open semantic view *without* changing the settings currently, and that seems weird to me. I mean, I hate it when software implicitly does something I don't necessarily want it to do like silently changes settings that are supposedly user-controlled.

Besides, having it change globally doesn't make much sense either, since whether you want to see it or not depends mostly on the project you're working on. So serializing semantic view visibility state makes much more sense than writing to global config.

russaa:
I did the following changes
 * removed direct DOM manipulation; instead using view-model & request re-rendering
   * am I correct in assuming, that etch does not require immutable properties/objects when updating the component state?
 * extracted a component for rendering the nodes
 * moved the semantic-view classes into a sub-directory
 * and also added a more complete serialization/deserialization

Currently I get a lint-warning that I should use triple-equality, where I really want to use double-equality -- do you have any idea how I can disable that warning for this single use/line?

lierdakil:
> where I really want to use double-equality

Uh... pretty sure it doesn't do what you think it does. `==` can do implicit type conversions, but for objects, it checks reference equality, exactly like `===`. Unless you pass a `string` and want to compare it with `[Object object]`, `==` is *not* your friend...

russaa:
I know the difference and usually I do use the triple-equality ... but in this case both values can be falsy, but I don't want to write out all explicit checks for all the possible combinations (boolean vs number vs object/null vs undefined vs empty-string)

lierdakil:
If I understand correctly, we're talking about comparing variables of type `NavigationTreeViewModel | null`. How can those be anything but `object` or `null`? `object` is never falsy, and it won't ever compare `true` to `null`. `null === null`. So... what's the problem?

russaa:
well, that's what the parameter-type specifies, but I am not 100% sure it could not be `undefined`, I would rather make a loose comparison than risk a failure  
(although in this case it would never result in an exception, still in principle I would like to have the option to use double-equality when I judge it appropriate)

but you have convinced me, trying to persuade the linter to ignore it is too much effort, I'll just change it

russaa:
sorry for getting of track.

Do you have any notes for improvement?

I think I addressed all mentioned concerns from before, but maybe not all in a satisfactory way(?)

lib/main/atom/views/outline/semanticView.ts
```diff
@@ -0,0 +1,64 @@
+import {NavigationTreeComponent} from "./navigationTreeComponent"
+import {NavigationTreeViewModel} from "./semanticViewModel"
+import {initialize} from "./semanticViewPane"
+
+export const SEMANTIC_VIEW_URI = "atomts-semantic-view"
+
+export interface SemanticViewOptions {
+  navTree?: NavigationTreeViewModel | null
```
lierdakil: :warning: 'Maybe' qualifier on `navTree` seems unnecessary here since it makes it `... | null | undefined`, which is less than optimal.

lib/main/atom/views/outline/semanticView.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export interface SemanticViewSerializationData {
+  data: {navTree: NavigationTreeViewModel | null}
```
lierdakil: :warning: `data` should probably be `SemanticViewOptions`

package.json
```diff
@@ -48,8 +48,18 @@
...
+    "showSemanticView": {
+      "title": "Show semantic view",
+      "description": "Show semantic view (outline) for typescript content",
+      "type": "boolean",
+      "default": false
     }
   },
+  "deserializers": {
+    "atomts-semantic-view/SemanticView": "deserializeSemanticView",
+    "atomts-semantic-view/SemanticViewPane": "deserializeSemanticViewPane"
```
lierdakil: :warning: There doesn't seem to be `deserializeSemanticViewPane` anywhere.

lib/main/atom/views/outline/semanticView.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export class SemanticView {
+  private comp: NavigationTreeComponent
+
+  public get element() {
+    return this.comp.element
+  }
+
+  constructor(public config: SemanticViewOptions) {
+    this.comp = new NavigationTreeComponent({navTree: (config && config.navTree) || null})
```
lierdakil: :warning: `config && ` is pointless here, since `config` is always defined.

lib/main/atom/views/outline/semanticViewPane.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.subscriptions.dispose()
+  }
+
+  async toggle(): Promise<void> {
+    if (!this.view) await this.show()
+    else await atom.workspace.toggle(this.view)
+  }
+
+  async show(): Promise<void> {
+    if (!this.view) this.view = new SemanticView({})
```
lierdakil: :warning: Since `SemanticView` could've been deserialized, need to check for that with `atom.workspace.itemForURI`. Otherwise a real chance we'll be creating a second one.

lib/main/atom/views/outline/semanticViewPane.ts
```diff
@@ -0,0 +1,74 @@
+import {CompositeDisposable} from "atom"
+import {SemanticView} from "./semanticView"
+import {Disposable} from "atom"
+
+class SemanticViewPane {
```
lierdakil: :art: `SemanticViewPane` isn't really a `Pane`, is it? Looks more like `SemanticViewController` at this point. Maybe a good idea to rename it to reflect its function.

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    return null
+  }
+
+  /**
+   * HELPER scroll the node's HTML representation (i.e. domNode) into view
+   *        (i.e. scroll the semantic-view's tree representation)
+   * @param  {Element} domNode the HTMLElement that should be made visisble
+   */
+  private scrollTo(domNode: Element): void {
+    const elem: any = domNode
```
lierdakil: :warning: Probably better to augment `Element` interface than cast to `any`

russaa:
thank you for this comprehensive review

I will go through the issues one by one an post another comment when I am through

lib/main/atom/views/outline/navigationTreeComponent.tsx
russaa: OK (changed in next commit)

lib/main/atom/views/outline/semanticViewPane.ts
russaa: OK (changed in next commit)

lib/main/atom/views/outline/semanticViewPane.ts
russaa: hmm, I will add a check in `show()`, if an instance is already present in one of the Panes

but still, I don't think the _leak_ is of big concern:  
first, it wouldn't happen often, and second, even though the `SemanticViewPane` / `SemanticViewController` will be bound in an additional `Disposable`, this `Disposable` should get garbage-collected without problems, since it is not referenced anywhere (cf. deserialization function `deserializeSemanticView` in `semanticView.ts`)

If it is still a concern for you, I could take additional measures to avoid creation of the additional `Disposable` -- should I?

lib/main/atom/views/outline/semanticView.ts
russaa: yes, I will change it

for explanation: that's just my paranoid nature -- there could still be `null` value, if someone really wanted to pass it in, e.g. using `null as any`, but I guess in that case it really should fail...

package.json
russaa: ahh, this is a leftover from something I reverted -- will be removed

lib/main/atom/views/outline/semanticView.ts
russaa: yes

lib/main/atom/views/outline/semanticView.ts
russaa: will be removed

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+          }
+        }
+      }
+      return true
+    }
+
+    return false
+  }
+
+  private whileRendering: {lastCursorLine: number | null} = {
```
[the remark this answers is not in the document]
russaa: well, it's a leftover from the previous implementation that I used as a starting point

I am not 100% sure I understand it completely, but I think it's purpose is to guard against the case that (a) currently there is no editor set and (b) there is no cursor set yet for the editor

I never tested, if (a) or (b) could really happen (since I just kept using the old code fragment)

lib/main/atom/views/outline/navigationNodeComponent.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  private renderNode(node: NavigationTreeViewModel | null): JSX.Element {
+    if (node === null) return <div />
+
+    const _root = this.props.root
+    if (!_root) return <div />
+
+    const selected =
+      (_root.selectedNode && this.isSameNode(node, _root.selectedNode)) ||
+      (!_root.selectedNode && _root.isSelected(node))
```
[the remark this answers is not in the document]
russaa: basically, `_root.isSelected()` is needed, if there's no node set as selected (yet):  
it calculates if the current cursor is within code-position that the node represents

this processing needs to be done, since it not enough to just check, if the cursor is within node's start- & end-range, as there could be a "more specfic" node within the start- & end-range that should be selected instead.

But I guess this is one the utility functions, that should get extracted

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+  private editor: TextEditor
+  public element: HTMLDivElement
+  private refs: {
+    main: HTMLOListElement
+  }
+  public selectedNode: NavigationTreeViewModel | null
+
+  private editorScrolling: Disposable
+  private editorChanging: Disposable
+  private activeEditorChanging: Disposable
```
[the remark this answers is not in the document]
russaa: will change

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    }
+    return null
+  }
+
+  /**
+   * HELPER scroll the node's HTML representation (i.e. domNode) into view
+   *        (i.e. scroll the semantic-view's tree representation)
+   * @param  {Element} domNode the HTMLElement that should be made visisble
+   */
+  private scrollTo(domNode: Element): void {
```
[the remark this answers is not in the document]
russaa: well, on MDN `scrollIntoViewIfNeeded()` is documented as [non-standard](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded)

and `scrollIntoView()` as [experimental](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) ... but `scrollIntoView()` is probably save to use(?)

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    this.whileRendering = {
+      lastCursorLine:
+        this.editor && this.editor.getLastCursor()
+          ? this.editor.getLastCursor().getBufferRow()
+          : null,
+    }
+    return (
+      <div class="atomts atomts-semantic-view native-key-bindings">
+        <ol ref="main" className="list-tree has-collapsable-children focusable-panel">
+          <NavigationNodeComponent {...{navTree: this.props.navTree, root: this}} />
```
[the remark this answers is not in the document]
russaa: will change this

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+        navTree.childItems = undefined
+        return
+      }
+
+      // TODO should there be a different sort-order?
+      //     for now: sort ascending by line-number
+      navTree.childItems.sort((a, b) => this.getNodeStartLine(a) - this.getNodeStartLine(b))
+
+      let child: NavigationTreeViewModel
+      for (child of navTree.childItems) {
```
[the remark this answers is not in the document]
russaa: will change 

(was just a matter of taste, as it feels weird using `const` in these cases -- it seems inconsistent to me, since the variable-reference is _not_ constant, and will get re-set in each loop)

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+   * @param {NavigationTreeViewModel} navTree
+   *            the NavigationTree that will be prepared for rendering
+   */
+  private prepareNavTree(navTree: NavigationTreeViewModel | null): void {
+    if (navTree === null) return
+    navTree.styleClasses = this.getIconForKind(navTree.kind)
+    const modifiersClasses = this.getClassForKindModifiers(navTree.kindModifiers)
+    if (modifiersClasses) {
+      navTree.styleClasses += " " + modifiersClasses
+    }
```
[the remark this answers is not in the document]
russaa: it's done here as a matter of "tree norminalization", before applying/rendering the new tree

but it can be moved into `NavigationNodeComponent` if necessary

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: `scrollIntoView()` is definitely safe to use.

And while `scrollIntoViewIfNeeded()` is a WebKit-specific extension, we can be pretty sure Atom won't move from WebKit anytime soon, and when it does, a lot of stuff would likely need fixing anyway, so I would argue it's safe to use *in Atom*. You can decide for yourself whether you're comfortable with this line of reasoning.

lib/main/atom/views/outline/semanticViewPane.ts
lierdakil: I guess this boils down to using the singleton pattern properly...

So here's an idea, implement the singleton pattern directly: make `SemanticView` constructor private, add a `create` static function and `instance` static private property. Then there would be no need to concern ourselves with whether `SemanticView` was created or not, since it wouldn't be possible to create a second instance *at all*.

Note though that it should be properly destroyed, i.e. `instance` should be set to `null`/`undefined` in destructor.

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    await etch.update(this)
+  }
+
+  private async loadNavTree(filePath?: string) {
+    filePath = filePath ? filePath : this.editor.getPath()
+    if (filePath) {
+      try {
+        const client = await clientResolver.get(filePath)
+        await client.executeOpen({file: filePath})
+        const navtreeResult = await client.executeNavTree({file: filePath as string})
```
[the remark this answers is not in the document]
russaa: was left over after it became unnecessary due to changed code -- will be removed

lib/main/atom/views/outline/semanticView.ts
lierdakil: I would argue that while there can indeed be anything anywhere since Javascript is weakly typed, we really should strive to catch that stuff before runtime, and thankfully TS makes it relatively simple. Otherwise, our code will become an unsightly mess of unnecessary checks everywhere, and that's just sad.

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: (b) isn't really possible, at least from my experience. (a) is not only possible, but it actually happens, since `render()` is called before the first call to `subscribeToEditor`. But I believe there has to be a more elegant way of handling it than this.

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: My point: why is assigning style classes to nodes is in "tree normalization" instead of in node presentation component?

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    this.selectedNode = null
+    await etch.update(this)
+  }
+
+  public async forceUpdate() {
+    await etch.update(this)
+  }
+
+  private async loadNavTree(filePath?: string) {
+    filePath = filePath ? filePath : this.editor.getPath()
```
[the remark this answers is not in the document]
russaa: you are right that the method is used only in one specific way a.t.m.

still, in an earlier implementation it was used differently (hence the optional argument) -- generally I am hesitant to remove code that can become potentially useful again

If you insist, I will change it, but in my opinion it should stay

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+    if (isEqual(navTree, this.props.navTree)) {
+      return
+    }
+    this.restoreCollapsed(navTree, this.props.navTree)
+    this.props.navTree = navTree
+    this.selectedNode = null
+    await etch.update(this)
+  }
+
+  public async forceUpdate() {
```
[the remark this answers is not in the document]
russaa: leftover from before etch -- will be removed

lib/main/atom/views/outline/navigationNodeComponent.tsx
```diff
@@ -0,0 +1,114 @@
...
+      (selected ? " selected" : "")
+
+    const domNode: JSX.Element = (
+      <li className={"node entry exanded list-" + classes}>
+        <div className="header list-item" on={{click: event => this.entryClicked(event, node)}}>
+          <span className={node.styleClasses}>{node.text || ""}</span>
+        </div>
+        <ol className="entries list-tree">
+          {node.childItems ? (
+            node.childItems.map(sn => <NavigationNodeComponent {...{navTree: sn, root: _root}} />)
```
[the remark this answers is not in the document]
russaa: will be changed

lib/main/atom/views/outline/navigationNodeComponent.tsx
```diff
@@ -0,0 +1,114 @@
...
+          ) : (
+            <div />
```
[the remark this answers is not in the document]
russaa: I did have `null` there, but at some point I got errors from the etch-renderer that there were invalid virtual node, so I changed it

When I change it back now, everything seems to work, so guess could leave it `null`...

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: The way I see it, `const` is an intent specifier. I.e. "`const`" means "I am not going to reassign this variable". I will concede that it's very javascript-y, and not at all consistent with other languages. But, personally, given a choice between code that makes sense in my head and code that is idiomatic, I always try to go for the latter, because more likely than not, I'm not the only person who's going to be hacking on that code, and idioms make life easier.

lib/main/atom/views/outline/navigationNodeComponent.tsx
```diff
@@ -0,0 +1,114 @@
...
+      _root.selectedNode = node
+    }
+
+    const classes =
+      (node.childItems ? "nested-" : "") +
+      "item" +
+      (node.collapsed ? " collapsed" : " expanded") +
+      (selected ? " selected" : "")
+
+    const domNode: JSX.Element = (
```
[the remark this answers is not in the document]
russaa: yes, you are right

(before, there was some code setting the selection after creating the DOM node, but that's not necessary anymore)

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
russaa: ... OK, I am not sure anymore:  
I tried moving it into the constructor of `NavigationNodeComponent` (before calling `etch.initialize(this)`)

... but then I get rendering problems: at some point the specific styling for the nodes, i.e. the icons, do disappear

Any ideas where this should be done, so that there will be no rendering problems?

.... the first few seconds everything seem all right (but this is actually the deserialized tree):
image

... but then it changes to this
image

looking at the HTML, I can see that the styling-classes vanish (or are not generated/set?), here the 
HTML code for one of the entries from the first screenshot:
```html
<li class="node entry exanded list-nested-item expanded selected">
  <div class="header list-item"><span class="icon icon-method modifier-public">refreshServiceUrl</span></div>
  <ol class="entries list-tree">
    <li class="node entry exanded list-item expanded">
      <div class="header list-item"><span class="icon icon-function">&lt;function&gt;</span></div>
      <ol class="entries list-tree"></ol>
    </li>
  </ol>
</li>
```

and here for the second:
```html
<li class="node entry exanded list-nested-item expanded selected">
  <div class="header list-item"><span class="undefined">refreshServiceUrl</span></div>
  <ol class="entries list-tree">
    <li class="node entry exanded list-item expanded">
      <div class="header list-item"><span class="undefined">&lt;function&gt;</span></div>
      <ol class="entries list-tree"></ol>
    </li>
  </ol>
</li>
```

lib/main/atom/views/outline/navigationNodeComponent.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  private toggleNode(node: NavigationTreeViewModel) {
+    // console.log("toggle " + !!node.collapsed + " -> " + !node.collapsed + " ", node) // DEBUG
+    node.collapsed = !node.collapsed
+  }
+
+  /**
+   * HACK detect click on collapse-/expand-icon
```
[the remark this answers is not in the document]
lierdakil: Hm. Okay then, I didn't think of the style guide. If it was just me, I'd probably convince myself that style guide is nothing more than a recommendation and just do what I deemed necessary anyway. But okay.

Still, after taking a closer look, this whole function seems to boil down to `!!node.childItems && event.target === event.currentTarget`.

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
russaa: OK, will be removed/changed

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: Try `NavigationNodeComponent.render`? I don't think this will have a performance impact. But if it would, then `NavigationNodeComponent.update` is as good a candidate as any.

If you're wondering, this is likely happening because Etch doesn't remove nodes if tags didn't change: https://github.com/atom/etch#keys

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  private async loadNavTree(filePath?: string) {
+    filePath = filePath ? filePath : this.editor.getPath()
+    if (filePath) {
+      try {
+        const client = await clientResolver.get(filePath)
+        await client.executeOpen({file: filePath})
+        const navtreeResult = await client.executeNavTree({file: filePath as string})
+        const navTree = navtreeResult ? navtreeResult.body! : undefined
```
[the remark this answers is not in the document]
russaa: OK, will be changed

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
russaa: ahh, of course, I also needed to update the style-classes in the `udpate` hook of the component -- OK, works now

lib/main/atom/views/outline/navigationNodeComponent.tsx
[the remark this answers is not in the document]
russaa: you are right, that is much simpler

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
russaa: I modified this a bit:  
extended the object to `PositionState` that hold the cursor-position and the `selectedNode` and passing it down the `NavigationNodeComponent`

-> so now, in `NavigationNodeComponent` the reference to the partent/root `NavigationTreeComponent` is only needed for `gotoNode(node)` upon a `click` event.

I am not sure, how to best get rid of this:  
I thought about an EventEmitter, but then you would need to link the EventEmitters of the `NavigationNodeComponent` so that they pass the event to their parent until root `NavigationTreeComponent` is reached.  
That doesn't seem very efficient.

Do you have any idea, how this could be accomplished?  
One possibility would be, to change `PositionState`, to also include such an EventEmitter, what do you think?

lierdakil:
Thanks for bearing with my nitpicks!

I took the liberty of going over the code once more and tweaking a few things. If you could look over 2538973 and comment on anything that looks questionable to you, that would be nice. I think we'll be good to merge after that.

russaa:
no problem -- it's good to get a review and being forced to rethink your decisions (or even really think about it the first time) ;-)

There is only one change, I am not so sure about:

you removed the `isSelected(node, _pos)` in `NavigationNodeComponent.render()`

While this allows you get rid of the `PositionState` etc. it does introduce some noticeable changes in the behavior of the NavTree representation:

for instance, if you watch the NavTree/SemanticTree representation while entering a new class/function/field, now, while typing, the selection oscillates, often enough ending up with the corresponding node not being selected.

I am not sure, if the code-simplification justifies this somewhat undesirable behavior(?)

russaa:
OK, I made some changes:
 
 * fixed to problem with the oscillating node-selection  
   (by setting the `selectedNode` when the `NavigationTree` is re-loaded/set)
 * extracted the dependencies/coupling of the `NavigationNodeComponent` from/with `NavigationTreeComponent` into interfaces:  
   this does not effectively decouple the node-component from the tree-component, but it makes the dependencies explicit

lierdakil:
Terribly sorry didn't get to it sooner. Thanks for fixing the bug I introduced, I was somewhat aware that there's something I forgot. Anyway, I think it's good to go, so merging. Thanks for your work!

Issue #1292, opened 2017-06-27
Make prettier script Windows compatible


russaa:
For plugin development:  
Currently, the plugin's prettier and test scripts (in `package.json`) cannot be run on Windows systems.

With some small modifications this can be salvaged, see pull request

PR #1293, opened 2017-06-27
Run prettier script via command-line arguments

russaa:

making the prettier script "compatible" with Windows, see issue #1292


scripts/prettier.ts
```diff
@@ -25,7 +26,9 @@ switch (process.env.ACTION) {
     }
     break
   default:
-    console.log("usage: ACTION=write|check prettier")
+    console.log(
+      "usage: prettier write|check\nor (*nix):\n       ACTION=write|check prettier\nor (Windows):\n       set ACTION=write|check&&prettier",
```
guncha: I think we should stick with one way to pass arguments. This script is not likely to ever require a lot of them so a regular positional argument works best on all systems, let's use that.

russaa:
OK, I removed the 2nd/original method for running the script by setting the environment variable `ACTION`


I am not quite sure, why the build fails:  
 1. the build fails on a file that was not touched: `lib/client/clientResolver.ts`
 2. the test script fails even before it comes to the part where I changed things, i.e. running the prettier script
     * I think it fails on the first part of the the test script, i.e. `tsc --noEmit -p ./lib`
  
I tested my changes (again and again) on a *nix machine, provoking some formatting errors, and it did fine checking&writing (node v7.0.0 and npm v4.5.0) ...

guncha:
That's a problem with the master branch, actually. It passed at the time I pushed it, but it doesn't look like we're pinning the dependencies so Typescript might have been updated which then is catching this error. I'll look into it and then we can merge this.

russaa:
ah, OK then -- I was starting to think, that something really weird was going on, either on my machine or the CI machine, but mystery solved :)

guncha:
@russaa the master is fixed now so you should merge it here and probably run `prettier` on the `prettier` script itself, heh 😄

russaa:
yep, done :thumbsup:

guncha:
👍

PR #1295, opened 2017-06-28
Remove conflicting typescript:quick-fix from keymap

sylvaindumont:
remove `typescript:quick-fix` from keymap, it does nothing and conflict `intentions:show`, resulting in codefixes unusable on linux or windows without editing keymap.

also document codefixes in readme


guncha:
Great catch, thanks! 👍

PR #1310, opened 2017-07-21
Atom 1.19 updates

lierdakil:
- [X] All compiled assets are included (atom packages are git tags and hence the built files need to be a part of the source control)

* Fixes "scheduleDidStopChangingEvent is not a function" (https://github.com/TypeStrong/atom-typescript/issues/1294)
* Fixes tooltip issue related to ye olden method of converting mouse event to buffer position not working anymore (#1324)
* Fixes `rootElement` deprecation warning (#1317)
* Fixes `ts-view`, in particular in 'find references' (#1303)
* Bumps dependencies
* Bumps minimal Atom version to 1.19.0

PR #1322, opened 2017-08-11
Fix autocomplete on Atom 1.19

anta-semenov:


lierdakil:
Basically the same fix is in https://github.com/TypeStrong/atom-typescript/pull/1310. Although you're right in using `debouncedEmitDidStopChangingEvent` instead of just `emitDidStopChangingEvent` -- which I missed initially.

Repo maintainers are apparently MIA, btw.

Also, привет, земляк!

lierdakil:
Fixed in 11.0.7 release.

Issue #1331, opened 2017-08-28
Typescript autocompletion significantly slow since atom 1.19 (incl. 1.20 betas) comparing to 1.18


deankevorkian:
On atom 1.18 versions, atom-typescript performs well (atom-typescript 11.0.6, Atom 1.18). Ever since I upgraded to 1.19 or 1.20 versions, atom-typescript performs slow. Tried reverting back to atom-typescript 11.0.6, but 11.0.6 doesn't work with atom >= 1.19. So I must use 11.0.7 which suffers from performance issues.

Any ideas about this? Anything you'd like me to check or details to add?
Thanks!

guncha:
What sort of slowness are you talking about? Is the autocomplete slow to open?

deankevorkian:
Yep, autocomplete suggestions are slow to open.

HitomiTenshi:
I had the same issue. Tried to switch because of this to VSCode but it doesn't have some features that I need.

Decided to downgrade to 1.18 for the meantime and autocomplete works fine again.

How it works in 1.18:
You start typing and autocompleted results already pop up. Almost instantly.

Slowness in 1.19:
You start typing and nothing. Only once you pause typing for 1 second it starts showing autocompleted results. Sometimes it doesn't even show that and I have to press CTRL + Space to force it to show results.

This delay kills productivity so hard I had to downgrade.

deankevorkian:
Same here, had to downgrade to 1.18 to enjoy it again. Maybe it's something cause by changes in atom's core?

GlenCFL:
I'm pretty sure that Atom 1.18 had a bug where onDidStopChanging would emit much more frequently than it should have, with it seemingly ignoring the timeout that should have been imposed on it. I'm not sure if atom-typescript's Autocomplete provider can be unhinged from onDidStopChanging, but I think that's the only way to regain that speed.

GlenCFL:
That would explain the speed difference in the current version of atom-typescript between Atom 1.18 and Atom 1.19+.

@lierdakil was there a particular reason to use debouncedEmitDidStopChanging over emitDidStopChanging in [this commit](https://github.com/TypeStrong/atom-typescript/commit/52811ad78569391805e9662e3f237a41ca54d43f)?

deankevorkian:
[...]

PR #1337, opened 2017-09-02
Reimplement typescriptBuffer::flush to speed up Autocomplete results.

GlenCFL:
This should resolve issue #1331.

The typescriptBuffer::flush() method now functions similarly to the way it did in atom-typescript v11.0.6, which results in autocomplete results being nearly instantaneous once again.


deankevorkian:
Tested on atom-1.20-beta5, works like a charm!
Could you please explain the changes? Why was the implementation changed to use `emitDidStopChangingEvent` in the first place?

GlenCFL:
@deankevorkian 
The root of the issue is that Autocomplete will call into this package as the user types, yet changes are unlikely to have been flushed out of the buffer at that point. The package author may then need a way to flush those changes out in order to provide up-to-date suggestions. Some Autocomplete providers can just get the current line from the buffer and work that way, but atom-typescript cannot.

atom-typescript has used scheduleDidStopChangingEvent, emitDidStopChangingEvent, and debouncedEmitDidStopChangingEvent within the TextBuffer class in order to flush out those changes. As you noticed, these are undocumented methods for that class, so they are subject to change from release to release of text-buffer (and thus Atom). The method used in Atom v1.18 in order to do so was scheduleDidStopChangingEvent. That method was removed in Atom v1.19, with the replacements being emitDidStopChangingEvent and debouncedEmitDidStopChangingEvent.

debouncedEmitDidStopChangingEvent would essentially have the Autocomplete provider wait for up to TextBuffer::stoppedChangingDelay milliseconds in order to flush out the changes and provide suggestions to the user, which is not ideal. In order to support emitDidStopChangingEvent, the changes in this commit were necessary, as otherwise we would be waiting for up to TextBuffer::stoppedChangingDelay regardless.

As a side note, the atom-typescript v11.0.6 implementation would temporarily set stoppedChangingDelay to 0 and flush out the changes that way when used in conjunction with scheduleDidStopChangingEvent. TextBuffer::emitDidStopChangingEvent accomplishes the same thing in a better way.

GlenCFL:
This needs additional work. I'll reopen this once it's completed.

GlenCFL:
Apparently we're leaking Disposables in the current release. The debounce prevented it from being a massive issue.

guncha:
@GlenCFL, thanks for investigating. Can you tell me more about the disposable leak?

GlenCFL:
@guncha
```ts
const {dispose} = this.buffer.onDidStopChanging(() => {
  dispose()
}
```
That just never actually disposes of itself, so we register a new callback to onDidStopChanging every single time we flush the buffer. Glancing through everything on the release commit again, it's not that big of an issue there.

It was a much bigger deal in this pull request due to its immediate nature, but it's fixed now. You would have thousands of callbacks firing after only a few minutes of editing.

deankevorkian:
@GlenCFL @guncha Could this PR get merged now? Autocompletion is extremely unhandy to use at the moment

guncha:
I'll test it today. @GlenCFL, if you can run the `prettier` on `./lib/main/typescriptBuffer.ts` and push the changes, that would be swell.

GlenCFL:
@guncha Already did.

This was the very first commit:
- Reimplement typescriptBuffer::flush to speed up Autocomplete results.

With this being the latest:
- Stop leaking Disposables.

The time stamps are wrong for the commits due to the time on my local machine being wrong. I had switched over from Linux to Windows, which results in the clock being wrong if you haven't applied a fix. If you want, then I can squash these down into a single commit, but there's no issue with an incorrect timestamp due to Git using commit hashes anyway.

guncha:
Oh, gotcha. I didn't know that Github uses the timestamp to order the commits instead of.. you know, the parent-child relationship of the commits. It should work the same once I merge to master.

lierdakil:
Sorry I didn't react sooner, but I didn't manage to find time to take a long hard look before now.

Using `debouncedEmitDidStopChangingEvent` was motivated by the want to avoid unnecessary events when typing, primarily -- I didn't realize it triggers on trailing edge. Personally, I barely notice the difference. I'm not using autocomplete-on-keypress though, so that might be why. Sorry about that.

Disposable leak is bad, but I'd like to point out I didn't touch that code, so we've been leaking disposables long before -- primarily because `Disposable.dispose` is not a bound method (https://github.com/atom/event-kit/blob/master/src/disposable.coffee#L31), so deconstructing and calling it with no object is an error (so frankly I doubt the dance around await is necessary, but oh well)

GlenCFL:
@lierdakil 
That's true. The issue is just in the destructured assignment. A cleaner way to do this would just be:
```ts
async flush() {
  if (this.changedAt > this.changedAtBatch) {
    await new Promise(resolve => {
      let sub = this.buffer.onDidStopChanging(() => {
        sub.dispose()
        resolve()
      })
      this.buffer.emitDidStopChangingEvent()
    })
  }
}
```
I really just went with the first thing that came to mind whenever I noticed the leak. I'll sneak it into a future PR if it doesn't appear there another way.

PR #1353, opened 2017-10-01
Setup specs

guncha:
I was going to work on the multiple buffer corruption and I thought it'd be good to test that behavior which led me down this rabbit hole of setting up a spec framework. This is the result. It uses Mocha and ts-node to compile spec files. As an example, I also added a spec to test that the package activates. 

### DIFF
--- a/spec/runner.js
+++ b/spec/runner.js
@@ -1,0 +1,39 @@
+// Atom already has the capability to require("foo.ts") files, but it's using typescript-simple package at version 1.0,
+// that is instead using Typescript 1.4. We want to use whatever version is specified in out package.json, so we kick out
+// the typescript-simple handler and install our own via ts-node.
+var path = require("path")
+
+if (require.extensions[".ts"]) {
+  // Because <insert reason>, the ".ts" extension property is read only and not configurable, so we have to replace the
+  // whole extensions object to get rid of it.
+  var extensions = {}
+
+  for (var ext in require.extensions) {
+    if (ext !== ".ts") {
+      extensions[ext] = require.extensions[ext]
+    }
+  }
+
+  require("module")._extensions = extensions
+}
+
+// Finally, register ts-node handler
+const {register} = require("ts-node")
+
+register({
+  project: __dirname
+})
+
+// Configure test runner and export the runner function
+var createRunner = require('atom-mocha-test-runner').createRunner
+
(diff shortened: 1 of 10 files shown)


spec/runner.js
```diff
@@ -0,0 +1,39 @@
+// Atom already has the capability to require("foo.ts") files, but it's using typescript-simple package at version 1.0,
+// that is instead using Typescript 1.4. We want to use whatever version is specified in out package.json, so we kick out
+// the typescript-simple handler and install our own via ts-node.
+var path = require("path")
```
lierdakil: We can use ES6 in Atom, so I would suggest `const` instead of `var`.

spec/runner.js
```diff
@@ -0,0 +1,39 @@
...
+
+if (require.extensions[".ts"]) {
+  // Because <insert reason>, the ".ts" extension property is read only and not configurable, so we have to replace the
+  // whole extensions object to get rid of it.
+  var extensions = {}
+
+  for (var ext in require.extensions) {
```
lierdakil: Can use `Object.assign({},require.extensions)` instead of this, and then `delete` from new object. A bit more compact, if nothing else, and I think `Object.assign` works somewhat faster on Chrome.

lierdakil:
https://github.com/TypeStrong/atom-typescript/commit/ab98ecef291c162eb4313e4048bcbd1a6f8227c1 contains proposed changes, cherry-pick that if you want to.

guncha:
Thanks for the review and tweaks. I'll include the typings via `spec/tsconfig.json`. You're right, they aren't necessary now, but the second you import a module from `lib`, it'll throw errors and I don't want some other contributor to have to spend time on setting that up.

Also, `mocha` globals are only available to code in `spec/` so I think it **should** be there.

lierdakil:
> Also, mocha globals are only available to code in spec/ so I think it should be there.

Uh... Pretty sure then you need to exclude `mocha` from `lib/` types, instead of adding it to `spec/` then?

> the second you import a module from lib, it'll throw errors

It won't. The only gotcha I could find is ad-hoc overrides and module definitions from `lib/typings/atom.d.ts` and `lib/typings/typings.d.ts` won't be visible without relevant references. Most of typings there are module-style, they don't contribute anything when `reference`d.

PR #1363, opened 2017-10-17
Nuclide codefix

lierdakil:

Basic support for `atom-ide-ui`'s code actions, as an alternative to intentions. This would allow to depend on `atom-ide-ui` exclusively, instead of current zoo of packages (`linter`, `linter-ui-default`, `hyperclick`, `intentions`)

This is basically a carbon copy of `CodefixProvider`, adapted slightly to `atom-ide-ui`'s API. There's probably an easier way to handle this mess, but I don't see it right now, and I'm running out of time.

Any input is greatly appreciated.

As a side note, this PR also updates README wrt dependencies.


lierdakil:
One thing I forgot to mention, I also bumped hyperclick provider from 0.0.0 to 0.1.0 (and changed the provider name as per [docs](https://github.com/facebook-atom/atom-ide-ui/blob/master/docs/definitions.md#hyperclick-service-api)), might solve some hyperclick-related issues

guncha:
I agree that his is a worthy goal and we should merge, but I'm bothered by the code duplication between `CodefixProvider` and `CodefixActionProvider`. The only differences are in the API. This can be easily abstracted out using your favorite compositional pattern. Here are some options:

1) Have both providers extend from a base provider that does the work
2) Extract the common code out into helper functions
3) Have one provider call the other and translate the responses
4) Have the same class implement both provider APIs

lierdakil:
Reasonable request. Done.

While at it, also changed some code in `applyFix`, namely switched from waiting for `save` event to using a promise directly.

A side note, `@types/atom` is horrifyingly outdated to the point of doing more harm than good. Switching to hand-rolled might be a better option at this point, and I do have a hand-rolled one [here](https://github.com/atom-haskell/typings) -- although it's rather rough around the edges, it's more or less up-to-date. Switching might be a lot of work, but I could probably do that. Let me know if you think it's a good idea.

lib/main/atom/codefix/codefixProvider.ts
```diff
@@ -110,4 +74,20 @@ export class CodefixProvider implements IntentionsProvider {
...
+      const {buffer, isOpen} = await this.getTypescriptBuffer(f.fileName)
+
+      buffer.buffer.transact(() => {
+        for (const edit of f.textChanges) {
+          buffer.buffer.setTextInRange(spanToRange(edit), edit.newText)
+        }
+      })
+
+      if (!isOpen) {
+        ;(buffer.buffer.save() as any).then(() => buffer.buffer.destroy())
```
guncha: You could `await` here for the `save()`. You won't need to fight with the typings that way.

lib/main/atom/codefix/codefixProvider.ts
```diff
@@ -60,33 +45,12 @@ export class CodefixProvider implements IntentionsProvider {
...
-                if (!isOpen) {
-                  buffer.buffer.save()
-                  buffer.on("saved", () => {
-                    buffer.buffer.destroy()
-                  })
-                }
-              })
-            },
-          })
+          results.push(formatCodeFix(fix))
```
guncha: Since you don't need to await or do anything else with the result you get from `formatCodeFix(fix)`, this method could have a return type of `Promise<protocol.CodeAction[]>` and the callers can take care of mapping over the array. It would make it a bit simpler.

lib/main/atom/codefix/codeActionsProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor(private codefixProvider: CodefixProvider) {}
+
+  async getCodeActions(
+    textEditor: AtomCore.IEditor,
+    range: TextBuffer.IRange,
+    diagnostics: Message[],
+  ): Promise<CodeAction[]> {
+    return this.codefixProvider.runCodeFix(textEditor, range.start, fix => ({
+      getTitle: async () => fix.description,
```
guncha: Neither of the two methods here need to be `async`.

guncha:
Thanks 👍

Yeah, `@types/atom` are pretty bad. If it's causing you issues when working on something, feel free to delete and start from scratch using the ones you have.

lib/main/atom/codefix/codeActionsProvider.ts
lierdakil: Actually they do, by the letter of the interface. I could use `Promise.resolve` of course, but I find this to be more succinct. Although you're right in that I should probably `await` `applyFix` below.

lib/main/atom/codefix/codefixProvider.ts
lierdakil: I've thought about that, but decided against, since we don't need to wait for one buffer to be destroyed to work on another.

lib/main/atom/codefix/codefixProvider.ts
lierdakil: Right.

lierdakil:
Hold off on a merge in any case, I think I found a bug I'd like to fix first.

lierdakil:
Okay, so apparently it's better to apply changes in reverse, otherwise multiple changes tend to trample over themselves. For example:
codefix-error

1f1d570 seems to work, but I didn't do any extensive testing.

lib/main/atom/codefix/codeActionsProvider.ts
guncha: Gotcha. Yeah, I also prefer using `async` over `Promise.resolve` even if it's a little less efficient.

lierdakil:
I've been using this for a while and didn't stumble upon any more issues, so I guess it should be good to merge. Which is exactly what I'm going to do now.

guncha:
🎉

PR #1364, opened 2017-10-23
Configurable autocomplete+ suggestionPriority

### LINKED ISSUE #1185: Turn off TypeScript snippets or prioritize JavaScript snippets
I've been struggling with this for a while now—is there a way to add in or prioritize JavaScript snippets for things like `console.log();` and `// TODO: `?  When TypeScript is selected as the language I get autocomplete snippets that I never use, like "interface LongRange" or "var localStorage" when I type "log".  If I add `console.log();` as a custom snippet (from the Atom menu —> Snippets on macOS) it gets sent to the very bottom of that list.  How can I make it default/at the top?  Is this a feature that needs to be implemented?

lierdakil:

Should close #1185.


yubaoquan:
When will this be merged?

guncha:
👍

PR #1367, opened 2017-10-30
Update typings

lierdakil:

So, long and short of it is this: I've switched to @GlenCFL's typings. Some stuff was wonky wrt possible `undefined` so I had to fix that.

While fixing stuff, I also found some things I really didn't like (like `var`), and so I also set tslint on it and fixed stuff *it* didn't like.

Also found some unused code, which I removed.

So long story short, I need someone who knows the codebase look over and give a go-ahead.

One change that isn't related to that is e5ef1c0 which removes progress bar from "emit" command. It's broken anyway (i.e. doesn't show progress) and it isn't hidden automatically for some reason, so I thought might be a good time to get rid of it.


GlenCFL:
`@types/atom` now has the majority of changes from `atom-typings`, with [DefintielyTyped#21334](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21334) bringing the remaining changes.

I'd suggest using `@types/atom` over `atom-typings` due to the special treatment `@types` gets from the TypeScript compiler. Pull requests to DefinitelyTyped should be faster now, with additional authors/reviewers (see [atom-typings#11](https://github.com/GlenCFL/atom-typings/issues/11)) being the way to speed up that process.

lierdakil:
Okay, so I'll just go ahead and merge this, lest it becomes stale. Feel free to shout at me and/or revert if something's broken.

Issue #1368, opened 2017-10-31
"needs to install" linter when atom-ide installed


lukescott:
I get the following dialog:

![dialog](https://user-images.githubusercontent.com/486224/32239988-5488249c-be29-11e7-96d7-6ca0bf4712eb.png)

I have atom-ide installed. The README says either atom-ide or linter, so it shouldn't be asking me to install linter if the other option is installed.

lierdakil:
Actually, readme you're looking at is for as-of-right-now unreleased version. Sorry about the confusion. Readme for the latest released version is on https://atom.io/packages/atom-typescript

But a fair point, we really should avoid this.

lierdakil:
@guncha, I'm wondering if we should default to `atom-ide-ui` actually if neither that nor `linter` is installed. Thoughts? I'll whip up a PR detailing proposed changes and rationale in a bit, but just in case asking here as well.

Thorski:
What action should I take when I get this dialog?   I have atom-ide-ui installed.   Do I need to install linter or not?

lukescott:
`atom-ide-ui` provides all the features that `linter` does. So you won't have to install `linter`. I believe `atom-ide-ui` provides additional features though. You can safely dismiss the dialog.

I click "No Thanks" every time I launch Atom to make sure this bug gets fixed eventually. You can click "Never" to not be bugged by it again.

lierdakil:
This should be fixed in v12. Thanks for your patience.

PR #1372, opened 2017-11-07
Atom deps update

lierdakil:

:rotating_light: **WARNING**:rotating_light: : This is based on `update-typings` branch, since working with old typings isn't much fun. I'm creating this PR on *that* branch for the sake of simplicity, but it means it requires a manual merge (since GitHub doesn't allow changing target branch after PR is created). It does not necessarily depend on new typings though, so it could be manually rebased onto current `master`.

So the main idea here is to check that any one of `atom-ide-ui` or `linter` is in the list of available (i.e. installed, although not necessarily active) packages. If neither is, prompt user to install `atom-ide-ui` via `atom-package-deps`, otherwise just continue with activation. Reason for `atom-ide-ui` instead of `linter`+friends is simplicity. `atom-ide-ui` is a single package, while `linter`+friends is not.

It seems like it's a lot more changes than it is due to indentation change. With `git diff -w` or `git log -p -w`, changes in `lib/main/atomts.ts` amount to (note a mix of word- and line-diff)

```diff
export {+async+} function activate() {
-  require("atom-package-deps")
-    .install("atom-typescript", true)
-    .then(() => {
+  const pns = atom.packages.getAvailablePackageNames()
+  if (!(pns.includes("atom-ide-ui") || pns.includes("linter"))) {
+    await require("atom-package-deps").install("atom-typescript", true)
+  }
```

I also changed target to `es2016` and changed `lib` accordingly for the sake of `Array.includes`, which is available at least since Atom 1.19, I believe. Could be more specific, but I'm pretty sure recent versions of Atom support es7 anyway.

Closes https://github.com/TypeStrong/atom-typescript/issues/1368


lierdakil:
Waiting on https://github.com/TypeStrong/atom-typescript/pull/1367. I really don't want to rebase this on `master` if I can avoid it.

lierdakil:
Actually found a bug introduced by this change. Code in `activate` used stuff initialized in `consume*` -- usually it's the other way around, but call order isn't guaranteed. Last two commits address that (first one moves config schema to package.json to avoid costly initialization when opening settings)

lierdakil:
Merged into master.

PR #1376, opened 2017-11-12
Automatically add import when autocompleting

noseglid:
It's a constant hassle manually adding imports for modules.
The changes here uses `source` from a CompletionEntry to
import the completed sybol.

Ideally, the modification of the source code should be handled
by `tsserver` too, but since it doesn't (yet) provide a feature
for it, we'll have to resolve doing it manually. As support in
tsserver for this is introduced, the `ImportManager` can hopefully
be phased out.

This works with relative includes (e.g. `../../core/Module`) and
using typescripts module resolution.

For instance, if you have a [compilerOption](https://www.typescriptlang.org/docs/handbook/compiler-options.html) which modified
paths:
```
baseUrl: "src",
paths: { "@/*": [ "*" ] }
```
It will automatically add imports like `import { Foo } from '@/Foo';`
rather than making a relative import.

Also adds two configuration which
  * Enabled/Disables the auto-import feature
  * Format the import string used


noseglid:
Here's a small gif of auto importing in action. Note how `SpaceMission` is imported using a custom module resolution, and `RocketShipType` is imported using a relative path.
ts-autoimport

lierdakil:
Hello. Thank you very much for this pull request. The idea is certainly interesting, although I find that codefix works fine too (especially with atom-ide-ui, but I think that code isn't released yet).

I will try to review the code next week, although I'm away from my main PC, on a dubious hotel Wi-Fi, so I can't make any promises. I have a couple requests before that though.

First, IIRC, we don't show autocompletion suggestions for modules not in scope (and it can become a bit of a nightmare for larger projects if we would), so could you explain how do you handle that?

Second, there's a branch/PR on this repo which has to happen before we merge anything else. See #1367 or `update-typings` branch. It also introduces linting (tslint). So if it's not too much trouble, could you rebase on that? I'll be back in December, and I'll merge #1367 first (if @guncha doesn't do it sooner), so I'd have to ask for a rebase anyway.

noseglid:
Since tsserver which shipped with 2.6, it includes completions from the full project, including `node_modules` (but in reality it seems that only `node_modules/@types` seem to be included.

I was unaware of the codefix and `intentions`. It seems there's a lot of overlap between the changes suggested in this PR and the ones provided by codefix. It also seem to support the module resolution.

As much as it pains me, I don't think this PR is worth moving forward with. The functionality is already in place (only that it's not applied when finishing an auto-complete).

Issue #1374, opened 2017-11-08
Installing atom-typescript does not disable built-in language-typescript


dotmagic:
I installed atom-typescript and was suprised some features didn't work (for example auto import). The reason is there are 2 different kind of Typescript Syntax Highlights, build-in package `language-typescript` and `atom-typescript` (this one). I believe if a user install atom-typescript, language-typescript should be disabled, otherwise it becomes very confusing. Reference https://github.com/TypeStrong/atom-typescript/issues/549#issuecomment-325934841 and #1344

lierdakil:
Well, that's certainly weird. And relatively recent. I haven't noticed that. Thanks for letting us know. 

Although, it's an open question whether we should disable built-in grammar. Maybe we should just drop our grammar and just use a built-in one. @guncha, any thoughts?

taylon:
I just started using Typescript and my experience has been pretty bad because currently we have atom-typescript, atom-ide/typescript-ide and language-typescript.

I chose to use atom-ide which means that I don't get the snippets and grammar from atom-typescript (which are pretty good) and language-typescript is currently really really bad.

I think that having a high quality built-in grammar is the best way to provide a good experience for whoever is using typescript, no matter if the user is in atom-ide, atom-typescript or none the experience would be great.

So IMHO moving the grammar from this package to language-typescript would be ideal, I would be happy to make the pull requests if you guys think it is a good idea and would help.

lierdakil:
Actually, it'd be rather hard for us to disable *any* package on installation, since we activate only when typescript grammar is used. Huh. This is a more difficult problem than I thought. Using Atom's bundled grammar seems more lucrative by the minute. @guncha, let me ask again: what do you think about dropping our grammar and using Atom's builtin instead?

taylon:
I personally think that using the builtin grammar is the way to go, I don't see the reason for duplication specially because the grammar might be useful without atom-typescript, someone using atom-ide for example will use the builtin grammar and won't use atom-typescript.

Also the grammar for Typescript in Atom (both builtin and atom-typescript) is currently really really bad compared to VSCode for example. If we concentrate the grammar in one place we have the potential to deliver improvements for all atom users independently of whether or not they are using atom-typescript.

lierdakil:
> someone using atom-ide 

That's just GUI. You still need backend, like atom-typescript or ide-typescript. So, irrelevant.

[...]

PR #1382, opened 2017-12-05
Switch to bundled grammar

lierdakil:

This is in response to https://github.com/TypeStrong/atom-typescript/issues/1374

Long story short, we should probably use bundled grammar. It's very nearly the same as ours (also converted from Microsoft's TextMate grammar), and there's little point in being stubborn and trying to maintain our own.

Scopes are also mostly the same, so the only difference, really, is activation hooks.


guncha:
Hehe, this is what I see when I search for `language-typescript`:

image

Is this because I'm behind the latest Atom?

guncha:
Yep, updated to 1.22.1 and it now shows as a bundled package (same repo, different version though). In principle, yes, it's a good idea to depend on it assuming Atom folks are going to keep it up to date.

lierdakil:
`language-typescript`'s last release was less than 24 hours ago (although since it's bundled, it's only *automatically* updated together with Atom), besides they have (somewhat minimalist currently) typescript support for atom-ide, so I expect they intend to keep it up-to-date for the foreseeable future.

On another note, thanks for the tip, `scripts/grammar.ts` (and relevant line in `package.json`) removed.

PR #1389, opened 2017-12-26
Keyboard tooltips

lierdakil:

This PR does several things, in addition to a command to show keyboard tooltips.

For one, it bumps dependencies in npm-shrinkwrap (because why not). Typings are also updated to latest applicable. Some changes are related to that, in particular, in `lib/main/atom/autoCompleteProvider.ts` and `lib/typings/atom.d.ts`.

`lib/main/atom/tooltipManager.ts` is rewritten to remove dependencies on `emissary` and `atom-space-pen-views` (with a notable exception of tooltip view). Since this was apparently the only dependency on `emissary`, that is removed from dependencies.

A small helper to effectively replace emissary is in `lib/main/atom/utils/element-listener.ts`

Actual tooltip command basically emulates mouse event by setting `clientX` and `clientY` based on last cursor buffer position.

As a side note, the command itself probably should be bound to `atom-text-editor`, not to `atom-workspace`. In fact, it makes no sense on `atom-workspace`. But the current (somewhat convoluted) way of dealing with commands only allows binding on `atom-workspace`. So... yeah, probably need to do something about that.

Sorry about messy commit history, I kinda forgot to make commits for a bit.


lierdakil:
Also, side note: command registrations are not disposed of properly. Probably should fix that at some point.

lierdakil:
@guncha I'm just going to merge this since it seems you're busy with other stuff. Feel free to revert and yell at me if you find it appropriate.

Issue #1332, opened 2017-08-29
[BUG] Import Name Constantly Misread


MrSnappingTurtle:
When importing modules, I often get a `...has no exported member...` error. When I hover over the module, it is looking for an odd spelling of the module:

![screen shot 2017-08-29 at 2 40 03 pm](https://user-images.githubusercontent.com/8148243/29845791-07f413a2-8cc9-11e7-9a53-83f343351809.png)

Notice it's looking for `MdToolbarModuleeludoM` instead of `MdToolbarModule`. Reopening the file fixes the issue.

lierdakil:
Most likely related to #1333. See if updating to Atom 1.19.5 fixes it for you.

MrSnappingTurtle:
@lierdakil I've updated to the latest. Still get the issue. Also, I noticed that it happens when I have two tabs open viewing the same file, if that's any help.

lierdakil:
Yes, that helps. That would be the reason this happens probably.

/cc @guncha, I'm pretty sure this happens because we're creating `TypescriptBuffer` for each `TextEditor`, while multiple `TextEditor`s can share the same `TextBuffer`. Hence, we get duplicate change events. Properly addressing this would entail a pretty major rewrite, so I'm not entirely sure it's something I should do -- at least not without asking first.

lierdakil:
Should be fixed in v12. Sorry about the long wait.

PR #1390, opened 2017-12-29
Cache constructed TypescriptBuffers by TextBuffers

lierdakil:

This is a fix for #1332 and similar issues. More of a band-aid, but this avoids major rewrites that would be required with other solutions.

Basically, it makes `TypescriptBuffer` constructor private and instead offers a `TypescriptBuffer::construct` "smart constructor" which returns an already-created `TypescriptBuffer` if it gets passed the same `TextBuffer` again. `TextBuffer -> TypescriptBuffer` correspondence is stored in a `WeakMap` keyed by `TextBuffer`, which allows garbage collector to do it's thing transparently. Overall, I find this to be a rather neat workaround.


lierdakil:
@guncha, I'm going to merge this as well. Also will try to make a release today. Feel free to yell at me if you think that's a bad idea.

PR #1393, opened 2018-01-10
Etch

lierdakil:

Long story short, this changes all views to use [etch](https://github.com/atom/etch) (Atom's react-like view framework) instead of jsx-render-dom or, heavens forbid, space-pen. For select-list-view I utilized [atom-select-list](https://github.com/atom/atom-select-list)

UPD: While working on this, I also switched to ES2017 target in 52392af. Core stuff is supported by Chrome 56, which Atom 1.23 is based around (and probably a few earlier, but I don't really care). Basically, only additional library stuff like `String.prototype.padStart` and atomics are not supported, and you need to enable those explicitly in tsconfig.
There are a couple caveats. One is that trailing comma in function definition/call isn't supported as well, but TypeScript never generates code like this. Another one is some `Proxy` and `Object` getter/setter-related stuff from ES2017 annex b is broken, but I don't think either we or TS use that.

Native async/await seems worth the trouble (because without those source maps aren't very useful frankly), but that could be argued. Not particularly attached to that, so will revert at first request.

TODO:
- [x] Forgot to set etch's scheduler to Atom's, will do later


lib/main/atom/components/miniEditor.ts
```diff
@@ -0,0 +1,65 @@
...
+      if (grammar) {
+        this.model.setGrammar(grammar)
+      }
+    }
+    this.model.scrollToBufferPosition([0, 0])
+  }
+
+  public async update(props: Props) {
+    this.element = atom.views.getView(this.model)
+    if (this.props.readOnly !== props.readOnly) {
```
lierdakil: TODO: check for undefined in `props.readOnly`, otherwise partial update is impossible

lierdakil:
Okay, so this is getting slightly out of hand. I'll just merge it without a review, lest it becomes either stale or overgrown.

Issue #1248, opened 2017-05-01
tsconfig.json


pvishnuteja:
unable to create tsconfig.json file

Neosublimation:
Me, too. The command vanished for me. It worked once.

BattleToads:
@guncha Any update on this?

guncha:
The update is that it's still missing 😄  You can run `tsc init` from the command line to initialize a `tsconfig.json` and that's what we should use if we wanted a command in Atom. This would work with different Typescript versions and would not need to be maintained as new options are added. Alas, no one has done the work for it yet.

Powell-v2:
Hey guys, I believe I can help you out with this one. What is the exact objective?

Neonit:
@Powell-v2 There once was a command for creating a default `tsconfig.json`. I pressed `Ctrl` + `Shift` + `P` (I think this is the default shortcut for getting the command input line in the Windows version), entered `tsconfig` and it suggested me a command for creating a default `tsconfig.json`. It vanished over half a year ago.

Powell-v2:
Thanks @Neonit  for the clarification 👍 

I've managed to add the command, in it's very basic form. There has to be a tab with any file opened, and based on its path, config file is generated in the same directory.

What are other considerations here?

guncha:
Here are some considerations:

1) The config file should be at the top of the project directory, if one is open.
1) We should also use the `tsc init` command to generate it to make sure it's future-proof. If you're just using a static string, it means you'll have to add any new options as they're added to the language.

Powell-v2:
Ok, gotcha! Already using the command, it's `tsc --init` btw ; )

nathan130200:
@guncha `tsc init`
> error TS5007: Cannot resolve referenced file: 'init'.
and `tsc --init`
> error TS5023: Unknown option 'init'

I have already latest version updated from npm.

PR #1395, opened 2018-01-16
Add initialize tsconfig command (#1248)

Powell-v2:

I've added requested functionality, however I wasn't entirely sure what to do in case of several open projects pertaining to one window. It goes without saying I'd be glad to iterate : )


lierdakil:
FYI, you can run tests locally:

```
npm install --only=dev
npm run test
```

Powell-v2:
Thanks for the tip, @lierdakil! Much appreciated.

lierdakil:
Thank you for the pull request. 

Unfortunately, the sheer volume of changed files suggests you've used a different version of typescript to compile assets. Try running `npm run build` in the project root.

More generally, while this works as a proof-of-concept, this is certainly not ready for production. First problem: it will run `tsc` from `PATH`, which might not even be there. A better idea: try to find `tsc` pretty much the same way we're trying to find `tsserver` here:

https://github.com/TypeStrong/atom-typescript/blob/31ffe23cad559dcc3c9be0698ee92a43c8b8f4b4/lib/client/clientResolver.ts#L113-L128

Should be defaulting to `tsc` we have bundled with the package, not to `PATH`.

Running synchronously is suboptimal. Using shell for execution is far from optimal. Using `cd` is especially troubling. Better idea: await a new `Promise<void>` wrapping `ChildProcess.execFile` which `resolve`s on success and `reject`s on failure. Specify `cwd` option in `execFile`.

Lastly, running around all open projects and executing `tsc --init` in each of those sounds like a *really* bad idea. Better idea: choose a project based on the currently-active editor's file path. Another idea: offer a selection dialog. If you opt for latter (selection dialog), I implore you to rebase on `etch` branch, since it's a) easier and b) we'll have to do that later anyway.

Powell-v2:
Thanks for quick reply and all the suggestions! I'm going to follow your instructions.

Could you help me to resolve a problem with `npm run build`? I get the following output:

```
> atom-typescript@12.0.0 build /Users/powell_v2/Downloads/atom-typescript
> tsc -p ./lib

lib/main/atom/autoCompleteProvider.ts(77,7): error TS2345: Argument of type '{ file: string; line: number; offset: number; prefix: string; includeExternalModuleExports: boole...' is not assignable to parameter of type 'CompletionsRequestArgs'.
  Object literal may only specify known properties, and 'includeExternalModuleExports' does not exist in type 'CompletionsRequestArgs'.
lib/main/atom/autoCompleteProvider.ts(81,47): error TS7006: Parameter 'entry' implicitly has an 'any' type.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! atom-typescript@12.0.0 build: `tsc -p ./lib`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the atom-typescript@12.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/powell_v2/.npm/_logs/2018-01-16T11_30_08_935Z-debug.log
```

lierdakil:
Seems like typescript version mismatch? Did you perchance forget to install package dependencies with `npm install` or `apm install` (both do the same thing)?

Relative to project root (apparently `/Users/powell_v2/Downloads/atom-typescript`?), `./node_modules/.bin/tsc` should exist and `./node_modules/.bin/tsc --version` command should output `Version 2.6.2`

Powell-v2:
I did follow steps listed in guide for contributors (it includes installing dependencies ... )

`./node_modules/.bin/tsc` exists, and I get exactly `Version 2.6.2`.

lierdakil:
Well, that's rather strange, to be perfectly honest. And, also to be perfectly honest, contributor's guide might be outdated, nobody took as much as a look at it since forever I'm afraid.

FWIW, `includeExternalModuleExports` should (hopefully) be defined in `./node_modules/typescript/lib/protocol.d.ts` on line 1215.

Anyway, you can just ignore it for now, `tsc` builds everything anyway. Worst-case I'll just have to recompile everything on my end, not a big deal.

Powell-v2:
I was able to successfully execute `npm run test`!

Still confused about all this pile of auto-generated files .. What ideally my commit should include?

lierdakil:
Any files you changed directly, plus anything `tsc` has put in `dist` (`dist/**/*.js` and `dist/**/*.js.map`)

Powell-v2:
Hi @lierdakil, I've consolidated everything into one commit, waiting for your review.

Btw I figured the issue I was experiencing yesterday - `atom-typescript` package update (to v12) was pending, didn't notice that! 😅

lierdakil:
General comment, I think it would make sense to `e.abortKeyBinding()` (which effectively lets something else handle the keypress if that was something that triggered the command) if `getPaths`/`getDirectories` returns empty array (so we have no project paths to initialize) *or* there's no active text editor.

Doesn't make all that much sense to abort if an editor doesn't have a path (most likely because it wasn't saved yet), a warning notification would be more appropriate.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,55 @@
+import {commands} from "./registry"
+import {execFile} from "child_process"
+import * as Resolve from "resolve"
+
+commands.set("typescript:initialize-config", () => {
+  return async e => {
+    const projectRootPaths = atom.project.getPaths()
+
+    if (!projectRootPaths) {
```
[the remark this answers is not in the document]
Powell-v2: Empty arrays are indeed never falsy, gonna change the condition as you proposed.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+  )
+}
+
+const initConfig = (tsc: string, projectRoot: string): Promise<void> => {
+  return new Promise<void>((resolve, reject) => {
+    try {
+      execFile(tsc, ["--init"], {
+        cwd: projectRoot,
+      })
+      resolve()
```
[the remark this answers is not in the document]
Powell-v2: True that!

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    if (!projectRootPaths) {
+      e.abortKeyBinding()
+      return
+    }
+
+    const editor = atom.workspace.getActiveTextEditor()
+    let currentlyActivePath
+
+    if (editor !== undefined) {
```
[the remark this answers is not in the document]
Powell-v2: I think `try/catch` is more descriptive and robust option here, taking into consideration your note below on `e.abortKeyBinding()` and  warning notification.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+      currentlyActivePath = editor.getPath()
+    }
+
+    const pathToTsc = await resolveModule("typescript/bin/tsc")
+
+    for (const projectRootPath of projectRootPaths) {
+      if (currentlyActivePath && currentlyActivePath.includes(projectRootPath)) {
```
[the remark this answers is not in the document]
Powell-v2: I'll take Atom API's route, should've researched docs better.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,55 @@
...
+        resolve(result)
+      }
+    }),
+  )
+}
+
+const initConfig = (tsc: string, projectRoot: string): Promise<void> => {
+  return new Promise<void>((resolve, reject) => {
+    try {
+      execFile(tsc, ["--init"], {
```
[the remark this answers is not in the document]
Powell-v2: Yep, that's right, I've added callback.

Powell-v2:
@lierdakil, corrections have been made as requested. However, I didn't completely understand which modifications for `resolveModule` you had in mind ... I didn't add *binaryName* argument to `resolveServer` either since I couldn't make it work in `initializeConfig.js` (returned value is `string | Server`, and we need it to be strictly `string`; probably I'm missing something fundamental here since I'm new to TypeScript). Please advise if possible, I'll gladly improve my code further 🤓

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+      currentPath = editor.getPath()
+
+      if (!currentPath) {
+        throw new Error("There is no active filepath available.")
+      }
+    } catch (e) {
+      console.error(e.message)
+    }
+
+    const pathToTsc = await resolveModule("typescript/bin/tsc").catch(() =>
```
lierdakil: :warning: Actually, after a closer look, we *really* need to pass `basedir` here, because otherwise `resolveModule` does pretty much the same as `require.resolve`. Using `resolveServer` looks even more lucrative then.

You don't really care if `resolveServer` returns `Server` since the interface *does* return path to executable, you can just ignore version by doing something like 
```ts
const pathToTsc = (await resolveServer(currentPath, 'tsc')).serverPath
```
note the `'tsc'` argument. Also probably a good idea to rename `resolveServer`, since it can resolve other stuff besides `tsserver`. I believe I mentioned already, that this `catch` can be moved to `resolveServer` too.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+        err => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve()
+          }
+        },
+      )
+    } catch (e) {
+      // error handling is done within callback since operation is async
```
lierdakil: :warning: `execFile` can throw some meaningful exceptions *synchronously* in case it can't create the process. So unless you want those to silently disappear, you'd want to `reject(e)` here as well.

lib/main/atom/commands/initializeConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+        throw new Error("There is no active text editor available.")
+      }
+
+      currentPath = editor.getPath()
+
+      if (!currentPath) {
+        throw new Error("There is no active filepath available.")
+      }
+    } catch (e) {
+      console.error(e.message)
```
lierdakil: :art: TBH, if all it does is `console.error`, then `try`/`catch` seems a bit pointless, it's more lines of code for the same effect as
```ts
if (!a) {
  console.error("message A")
  return
}
// ...
if (!b) {
  console.error("message B")
  return
}
```

If it would do some error handling besides logging to console, it would make much more sense. Which it can, by the way, doing `e.abortKeyBinding()` as appropriate, or showing a nice user-friendly message that something went wrong (via `atom.notifications.addWarning` for example)

Note that you can differentiate different errors by their message, f.ex.
```ts
try {
    // ...
    throw new Error("ENOPROJECT")
    // ...
    throw new Error("ENOEDITOR")
  } catch (e) {
    switch (e.message) {
      case "ENOPROJECT":
      case "ENOEDITOR":
        e.abortKeyBinding()
        return // we handled those, yay
      default:
        // required because you can never be sure what is actually thrown
        if (e.stack) {
          // because we want to be sure it has stack,
          // otherwise weirdness will ensue
          atom.notifications.addFatalError("Something went wrong", {
            detail: e.message,
            dismissable: true,
            stack: e.stack,
          })
        } else {
          atom.notifications.addError("Something went wrong and we don't know where", {
            detail: e.toString(),
            dismissable: true,
          })
        }
        return
    }
  }
```
(note: this is very general, here you can safely assume that `e` is instance of `Error` and has `stack` defined)

lib/client/clientResolver.ts
```diff
@@ -98,7 +98,7 @@ export class ClientResolver extends events.EventEmitter {
 }
 
 // Promisify the async resolve function
-const resolveModule = (id: string, opts: Resolve.AsyncOpts): Promise<string> => {
+export const resolveModule = (id: string, opts: Resolve.AsyncOpts = {}): Promise<string> => {
```
lierdakil: :warning: after taking a closer look, we really *don't* want this `= {}` here.

lib/main/atom/commands/initializeConfig.ts
Powell-v2: I think I finally got it!

lib/main/atom/commands/initializeConfig.ts
Powell-v2: Good to know.

Powell-v2:
Another commit inbound. I've done my best to follow your instructions. Thanks a lot by the way for your time, patience, and detailed replies - all of it is highly appreciated! 🙏

lierdakil:
Okay, at this point it's much easier for me to go ahead and fix some stuff mostly related to code style by hand. I hope you don't mind too much that I kinda take over your work, but it would literally take significantly more key presses to explain than to fix :)

lierdakil:
I think I'm just going to merge this now. Thank you for your work!

Issue #679, opened 2015-10-22
Go To Declaration should center cursor


mjohnsonengr:
When I press F12 to go to a declaration, the cursor is moved to the destination line and the view scrolls such that the destination is very close to the bottom of the visible area.  It would be nice (and maybe others would agree) if the view adjusted so that the destination line shows up at approximately the middle of the visible area.

My justifications for this suggestion are two-fold, although both reveal that I may be either biased or spoiled.  Discussion will determine which, if either holds true.
1. I am used to Visual Studio which does this (other users may also be migrating to Atom from the same situation)
2. I flipped my monitor 90 degrees so it is in portrait orientation.  It displays about 110 lines of code on the screen, so I tend to lose track of the cursor when it is at the bottom of the screen.

Thoughts?

mjohnsonengr:
I missed the point that may seal the deal.

About 80% of the time that I use Go To Declaration, I'm navigating to a method, and will most likely be looking at the following 10-20 lines very soon after this action.

I keep thinking that other people will probably prefer this differently, which indicates that this might be a candidate for a customization option.  Correct me if it already is and I missed it.

basarat:
We just ask atom to bring the line in view : https://github.com/TypeStrong/atom-typescript/blob/c0d11587e70b54272d7b5ad4497dd5fa35b88fe1/lib/main/atom/gotoHistory.ts#L17-L23

If atom provides an API will will use it ... if not don't feel like writing / maintaining it. Sorry :rose: but a good suggestion nonetheless.

basarat:
Leaving open in case someone files the API endpoint :rose:

dsifford:
I'd love to be able to handle this and send a PR, but I really have no idea where to even begin (atom's api is extremely foreign to me).


That being said, I think these two resources might be handy for those who are able to decipher the api and the structure of this project....


How is was done in atom bookmarks: https://github.com/atom/bookmarks/pull/45/files
The (possibly) relevant api method: https://atom.io/docs/api/v1.7.3/TextEditor#instance-scrollToCursorPosition

Hope that is at least somewhat helpful. 👍

lierdakil:
Released as v12.2.0 :tada:

PR #1402, opened 2018-02-03
Center cursor after open in goToDeclaration

thepelkus-too:
Per #679, when going to a declaration, the cursor would often be at the
bottom of the screen. The cursor should now be moved to the middle of
the screen per a call to TextEditor#scrollToCursorPosition once the open
promise resolves.

Thanks to @dsifford for pointing out the docs for the scroll method.


lib/main/atom/commands/goToDeclaration.tsx
```diff
@@ -5,11 +5,15 @@ import * as etch from "etch"
...
 const prevCursorPositions: FileLocationQuery[] = []
 
-function open(item: {file: string; start: {line: number; offset: number}}) {
-  atom.workspace.open(item.file, {
+async function open(item: {file: string; start: {line: number; offset: number}}) {
+  await atom.workspace.open(item.file, {
     initialLine: item.start.line - 1,
     initialColumn: item.start.offset - 1,
   })
+  const activeEditor = atom.workspace.getActiveTextEditor()
```
lierdakil: Do not assume that active editor is the one you've opened. Take the result of `atom.workspace.open` instead, it's guaranteed to return what it's opened.

lierdakil:
Thanks for the pull request, by the way!

PR #1406, opened 2018-02-26
Correct link to change log

fedde-s:
The link pointed to a non-existent file in a different location,
I presume that it was simply outdated.

- [ ] All compiled assets are included (atom packages are git tags and hence the built files need to be a part of the source control)  <-- Can I assume that this isn't relevant for a documentation-only fix?


fedde-s:
If this commit doesn't fit the workflow somehow and you want to make your own commit, don't worry about attributing me for this four-byte documentation fix 😄

lierdakil:
Thanks.

Issue #1219, opened 2017-03-22
Symbols-view integration is missing


diminutivesloop:
Before the 11.0.1 update I could access Typescript symbols via the symbols-view package's toggle-symbols and toggle-project-symbols commands. This was a particularly convenient feature for navigating large codebases and large files, so I'd really appreciate if it can be fixed or re-implemented.

guncha:
Yeah, it's one of the things that didn't make it over, sorry. I don't even know how it could be reimplemented since we don't have direct access to the Typescript AST anymore, but I'm sure it's possible. If there's enough demand for it, might be worth it.

dominic31:
+1, would be great to get this back.

robofish:
Would also be lovely to get back support for `symbols-view:go-to-declaration` / the "Go To Declaration" in the context menu 😀

manuelnaturalcycles:
@guncha can you point me in the direction needed to implement this? i would like to give it a try because this seriously breaks atom for me. i cant really work without go to declaration keybind, i work in a massive typescript project.

thanks and sorry for bothering you.

guncha:
@manuelnaturalcycles, we have a `Go to declaration` command that is bound to `F12`. I guess there's a `Symbols view: Go to declaration` that doesn't do anything. I should get rid of that. It's not in the context menu though, is that the problem?

If you mean implementing Symbols view, it's going to be hard, because AFAIK there's no way to access all the symbols for a given file via the `tsserver` API. You could ask for `quickInfo` for every location in the file, but that would be madness.

manuelnaturalcycles:
@gunchaI i see.
i hate myself for not having the knowledge to approach this. 
i remember having this keybind in atom:
`'atom-workspace atom-text-editor:not([mini])':
  'cmd-g': 'symbols-view:go-to-declaration'`
and it worked. i wondered if it was possible to do some work without going too deep to bring that functionality back. but it seems to be pretty deep into things i dont even know of :(

andregomars:
+1, need native symbol view support of .ts files

mashleyttu:
Would love to have this feature back. Used it consistently. :)

russaa:
maybe the ["navtree" command](https://github.com/Microsoft/TypeScript/blob/master/lib/protocol.d.ts#L1623) from the `tsserver` API could be used for creating a symbols-list/-tree.

A quick test with a small example:
```typescript
export class Greeter {
    greeting: string;
    constructor(message: string) {
[...]

PR #1407, opened 2018-02-26
Re-Adding Symbols View (#1219)

russaa:

Re-adding `symbols-view` for project- and file-symbols  
-> see discussion in #1219

This PR is based on the `symbols-view` system package with some small modifications:
 * code integration/modifications for including it here as TS resources
 * replaced `ctag` parser with `tsserver` commands `navTree` (for file symbols) and `navto` (for project sybmols)

It also includes keyboard shortcuts for overriding the `symbols-view` shortcuts if a TS file editor window is focused (i.e. `ctrl-/cmd-r` and `ctrl-/cmd-shift-r`).


lierdakil:
Sorry, I changed a bunch of architecture-related stuff on `master, could you rebase?

Most notably, feature registration moved from `atomts.ts` to `lib/main/plugin-manager.ts` (since it's easier to manage lifetimes that way), and command registration changed up a bit in `lib/main/atom/commands`.

Overall, this feels like *way* too much code for something as simple as "show symbols for file/project". At least views seem to be crazily overengineered. It would seem to me, in most cases, a simple call to `selectListView` from `lib/main/atom/views/simpleSelectionView.tsx` would suffice. Maybe let's try to simplify because I *really* don't want to end up maintaining this mess.

russaa:
As I mentioned:  
this is basically a copy of the files from `symbols-view`.

So, in my understanding, maintaining would mean:
 * diff with `symbols-view` and transfer changes
 * maintain additional/custom code (mainly with regard to using `tsserver` and converting the results to a format that `symbols-view` "understands)

So rather than re-implementing this, I would prefer to make the changes more explicit (where that is not already done), so that the "diff & transfer from `symbols-view`" portion of maintenance is painless

Of course the require re-formatting in `atom-typescript` makes this a bit more difficult, but it's still pretty straight forward.  
Also: `symbols-view` seems to be pretty stable, i.e. is not updated too often.

Ideally at some point, the `symbols-view` package may be opened up to allow different providers (other than ctag-parsing), at which point these views could be removed again.

russaa:
... still I think, it would be better to keep the _copy_ as is:  
 * regardless what reason, updates from `symbols-view` should be required very rarely
 * and you'd only really need to update/transfer changes, if something breaks or the UI significantly changes in `symbols-view`

-> so I would maybe extract all changes from the current files (i.e. keep the copies as close to the original as possible, except for the necessary changes due to conversion to Typescript) 
-> include the changes by extending the original/base classes

and make formatting exceptions for the original/base class files

Then in almost all instances, you'd only look at the modification/extending files, and leave the original/base class files as is.

I had a quick look at what could be removed: it would be mostly stuff to watch for file-changes (which would trigger ctag-parsing/updates) and loading the ctag-caching files ... I did not find too many places with code for this, so I think, that --if you want to keep the `symbols-view` behavior-- you need to keep most of the code.

Let me know what you want to do:  
if you decide to pursue the option of "keeping the copies", I could make the changes that I suggested above (extract the modifications to extra files, and keeping the original/copied files as close to the `symbols-view` implementation files as possible).

lierdakil:
I might have gotten a little carried away with this... Anyway, it seems most of what could be done has been done here. So I'll merge this if that's okay with you.

russaa:
looks all good to me 👍 

_I mean to symbols-view part -- to be honest, I did not look too much into the other stuff regarding position history etc_

PR #1408, opened 2018-02-28
[experimental] Symbols view

lierdakil:

@russaa, this is based on #1407 

One big change I made is instead of using `navTo` for project navigation, I just run `navTree` for all project files (except those in `/node_modules/`, but the exclusion algorithm is stupid and needs some work). That's not necessarily because I think it's better, just because I didn't have time to fumble with streaming loading.

Oh, yes, and I moved getting navTree into `TypescriptBuffer`.

This is by no means perfect emulation of `symbols-view`, but the code ended up being *a lot* more straightforward IMO, and I would argue that we don't *want* the perfect emulation of `symbols-view`, since it refers to `ctags` more than once, which can be confusing.

The only features missing compared to `symbols-view` at the moment are `symbols-view:return-from-declaration` and highlighting of the matched prefix in select-view.

The former we have already, actually, so we just need to hook new code into that.

The latter I couldn't be bothered with at the moment, but it *should* be rather straightforward.

Anyway, let me know if you think this seems somewhat reasonable.


russaa:
overall, it looks good to me -- and much less code than my _copy_ approach

... although I am not sure about replacing `navto` with `navTree`:  
 1) it takes quite some time to initialize when searching for project-symbols (each time), during which the UI is unresponsive
 2) it excludes `/node_modules/` which were included when using `navto`

Also:  
when trying this, I sometimes get an error `Error processing request. No Project.` (but that can probably be remedied)

as for 1):  
in difference to using `navTree` on all files, `navto` is quite fast, even though there is a new request on each keystroke/typed letter -- maybe this can be remedied by caching the results when running `navTree` on all files (but you'd need to watch for changes and update the cached data).  
A possible drawback using `navto` (and advantage for `navTree`): `navto` does not show a list for an empty search-string, which the approach using `navTree` on all files does. It's probably a matter of taste, if results for an empty search-string is important or not.

as for 2):  
Not getting results for `/node_modules/` may not necessarily be unwanted, as it may be matter of taste, if you want results from there -- I personally prefer to have it included, since sometimes I like to look up the libraries etc. if I need to know, how I can use a method etc. and documentation is lacking


-> so for the decision --to replace `navto` with `navTree`-- it probably depends on:  
if you consider the issues (see above) important or not and, if they are important, how much effort it would take to address these using the `navTree` approach

russaa:
... if you consider it viable/worth trying, I could look into using `navto` for project-symbols again

lierdakil:
> it takes quite some time to initialize when searching for project-symbols (each time)

Actually, getting `navTree` for all files is rather fast and does *not* block the UI thread. Rendering all those items *is* a performance bottleneck, however.

> it excludes /node_modules/

This was a [deliberate choice](https://github.com/TypeStrong/atom-typescript/pull/1408/files#diff-9c66dc4e7d71e9c27e5f9f7d92e840b5R18) on my part, actually, mostly for performance reasons (`atom-select-list` becomes noticeably sluggish after a couple thousand items)

---

Overall, yes, using `navTo` might be worth it in the end, since a complete list of all project symbols is not very useful anyway. And since you did most of the work in #1407 anyway, I went ahead and reinstated that.

I did, however, rewrite parsing functions because those were somewhat ugly.

Also limiting the maximum number of results from `navTo` to 1000, because having even 1000 items is way beyond reasonable IMO -- I don't think anybody will want to scroll through those.

If you could look into adding matched part highlighting to `lib/main/atom/views/simpleSelectionView.tsx`, that would be great.

Also, keep an eye out for `Error processing request. No Project.`, if it's not too much trouble, because I have no idea how or why that might happen -- didn't run into that myself.

russaa:
OK, good.

I'll see to the highlighting

The `Error processing request. No Project.`-error was thrown for `navTree` requests -- I think this might have been for Typescript files that were in the project, but not within the sources-folder(s) that the `tsserver` was watching/compiling.

If you want, I could investigate this in more detail -- but using `navto` should solve this problem anyway, since it will "ignore out-of-project" files automatically.

russaa:
OK, I added the highlighting (details in the PR #1409)

let me know, if I should change something

_(as expected, I did not see any errors `Error processing request. No Project.` anymore)_

lierdakil:
I'm moving this back to #1407 to avoid pointless pull requests.

PR #1409, opened 2018-02-28
[experimental] Symbols view with highlighting

russaa:

added highlighting to #1408 

Additional noteworthy changes:
 * added CSS class `atom-typescript` to the root-element in `simpleSelectView` (for enabling specific/targeted styling of the `SelectListView`)
 * renamed file `fileSymbolsTag.ts` to `symbolsTag.ts` (since the `Tag` class is used for both file- and project-symbols)


lib/main/atom/views/simpleSelectionView.tsx
```diff
@@ -51,6 +52,7 @@ export async function selectListView<T>({
         didChangeQuery,
         emptyMessage,
       })
+      select.element.classList.add("atom-typescript")
```
lierdakil: There is a `itemsClassList` property that's designed for this. Probably a good idea to use that...

lib/main/atom/views/symbols/utils.ts
```diff
@@ -25,3 +27,42 @@ export function deserializeEditorState(editor: TextEditor, {bufferRanges, scroll
...
+  query: string,
+  offsetIndex: number = 0,
+): QueryMatch[] {
+  let lastIndex: number = 0
+  let matchedChars: string[] = [] // Build up a set of matched chars to be more semantic
+  const queryMatches: QueryMatch[] = []
+
+  const matches: number[] = match(name, query)
+  let matchIndex: number
+  for (matchIndex of Array.from(matches)) {
```
lierdakil: Unnecessary `Array.from`?

lib/main/atom/views/symbols/utils.ts
```diff
@@ -25,3 +27,42 @@ export function deserializeEditorState(editor: TextEditor, {bufferRanges, scroll
   editor.setSelectedBufferRanges(bufferRanges)
   editorElement.setScrollTop(scrollTop)
 }
+
+// extracted/adapted from symbols-view package (symbols-view.js::SymbolsView.highlightMatches)
+export function highlightMatches(
+  name: string,
+  query: string,
+  offsetIndex: number = 0,
```
lierdakil: What is the purpose of `offsetIndex`? It seems like it's not used anywhere.

lib/main/atom/views/symbols/fileSymbolsView.tsx
```diff
@@ -16,9 +17,12 @@ export async function toggle(editor: TextEditor, deps: Deps) {
     }
     const tag = await selectListView({
       items: generateFile(filePath, deps),
-      itemTemplate: ({name, position}) => (
+      itemTemplate: ({name, position}, ctx) => (
         <li class="two-lines">
-          <div class="primary-line">{name}</div>
+          <HighlightComponent
+            styleClass="primary-line"
+            matches={utils.highlightMatches(name, ctx.getFilterQuery())}
```
lierdakil: I'm pretty sure we're not going to use `HighlightComponent` with any matching engine besides `highlightMatches`. Maybe better have `name` and `query` attributes on `HighlightComponent` and push `hightlightMatches` call inside?
Also while at it, I would prefer to have `HighlightComponent` to render as an inline element to have more flexibility, since it might be a good idea to reuse it in 'go to declaration` etc. So
```tsx
<div class="primary-line">
<HighlightComponent ... />
</div>
```
is wordier, but is more flexible then what you have here.

lib/main/atom/views/symbols/highlightComponent.tsx
```diff
@@ -0,0 +1,30 @@
...
+    await etch.update(this)
+  }
+
+  public async destroy() {
+    await etch.destroy(this)
+  }
+
+  public render(): JSX.Element {
+    return (
+      <div class={this.props.styleClass || ""}>
```
lierdakil: As mentioned above, probably a better idea to render this as an inline element, so `span` instead of `div` seems reasonable.

Also, while here, setting `class` property to `""` will create an empty `class` attribute on the DOM element. Setting `class` property to `undefined` will not create `class` attribute at all.

While at it, might as well rename `styleClass` to just `class` for the sake of consistency. The trick is to use quotes in the type definition.

russaa:
OK, I think I did all the changes, could you review again?

lierdakil:
Thanks. I'll just merge this right now. We don't really need `itemsClassList` to be configurable because a stylesheet kinda depends on it being `atom-typescript`, but it's easier to fix on my end.

PR #1417, opened 2018-04-12
Adding refactor commands

### LINKED ISSUE #1401: Use tsserver refactor capabilities
Hi!

TypeScript `tsserver` has commands for refactoring, AFAIK, the commands are `GetApplicableRefactors` and `GetEditsForRefactor`.

Is there any plan to include these in `atom-typescript`? It would be a very nice feature.

InExtremaRes:

It uses the `getApplicableRefactors` with the current selection and, if there is a positive response, show the list of refactors in a `simpleSelectionView`. When the users pick one of them it sends a `getEditsForRefactor` command and perform the edits in the buffer.

There are at least two remaining tasks:

- Show a message to the user when there is no applicable refactors (now it just closes the selection).
- Call a `rename` command after the edits.

I don't know if there is a way to call another command from inside the callback of `addCommand` itself.

Nonetheless the command works (I think :stuck_out_tongue:)

Fixes #1401


lib/main/atom/commands/refactorCode.tsx
```diff
@@ -0,0 +1,131 @@
...
+  if (responseEdits.body === undefined) {
+    e.abortKeyBinding()
+    return
+  }
+
+  for (const edit of responseEdits.body.edits) {
+    await deps.withTypescriptBuffer(edit.fileName, async buffer => {
+      buffer.buffer.transact(() => {
+        for (const change of edit.textChanges.reverse()) {
+          editor.setTextInBufferRange(spanToRange(change), change.newText)
```
lierdakil: Uh... I'm pretty sure it was supposed to be `buffer.setTextInRange`?

lierdakil: Also, while at it, this (or very similar) code is used at least in three places already. Maybe it's a good idea to refactor that into a function along the lines of `async applyEdits(edits: FileCodeEdits[]): Proimse<void>`

lib/main/atom/commands/refactorCode.tsx
```diff
@@ -0,0 +1,131 @@
...
+  client: TypescriptServiceClient,
+  range: protocol.FileRangeRequestArgs,
+) {
+  const responseApplicable = await client.execute("getApplicableRefactors", range)
+  if (responseApplicable.body === undefined || responseApplicable.body.length === 0) {
+    return []
+  }
+
+  const actions: RefactorAction[] = []
+  responseApplicable.body.forEach(refactor => {
```
lierdakil: TBH, I think `for(const ... of ...)` would be more readable here.

lib/main/atom/commands/refactorCode.tsx
```diff
@@ -0,0 +1,131 @@
...
+              label={`${item.refactorDescription}: ${item.actionDescription}`}
+              query={ctx.getFilterQuery()}
+            />
+          </li>
+        )
+      },
+      itemFilterKey: "actionDescription",
+    })
+
+    if (selectedAction !== undefined) {
```
lierdakil: For the sake of consistency, I would prefer
```ts
if (!selectedAction) return
await applyRefactors(...)
```

lib/main/atom/commands/refactorCode.tsx
```diff
@@ -0,0 +1,131 @@
...
+      startOffset: selection.start.column + 1,
+      endLine: selection.end.row + 1,
+      endOffset: selection.end.column + 1,
+    }
+
+    const actions = await getApplicableRefactorsActions(client, fileRange)
+
+    if (actions.length === 0) {
+      // TODO Show a "no applicable refactors here" message
+      e.abortKeyBinding()
```
lierdakil: Really shouldn't `abortKeyBinding` here if we're showing the message. The idea of `abortKeyBinding` is to behave as if this handler wasn't executed in the first place (so that other handlers might do their work, e.g. if different commands are bound to the same key). If you're showing the message, that's a side-effect that shows that the handler was executed, so it really makes little sense pretending it wasn't.

InExtremaRes: Oh, sure. That block should be replaced with that of show message.

lib/main/atom/commands/refactorCode.tsx
InExtremaRes: https://github.com/TypeStrong/atom-typescript/blob/b4fb237e1934ba1e7373dbdc1c7c3aeba0676130/lib/main/atom/codefix/codefixProvider.ts#L56-L65

That is one of the places you mention of, but it is the same `buffer.buffer.transact` sentence. In case of refactor that code away in a function, which would be the expression to keep?

lib/main/atom/commands/refactorCode.tsx
InExtremaRes: Since there was a [strict-boolean-expressions](https://github.com/TypeStrong/atom-typescript/commit/f5ec4f162524365684f91bdb342188cc152b98c9) commit, wouldn't be preferred to use
```
if (selected === undefined) ...
```
instead?

lierdakil: We have `allow-null-union` and `allow-undefined-union` enabled. So unless the value can be falsy some other way, boolean coercions like this are allowed.

The reason is simple: non-null objects are never falsy, and it's usually obvious from context whether something's an object or a boolean.

I mean, I'm all for clarity, but there's clear code and then there's code that's unnecessarily verbose.

lib/main/atom/commands/refactorCode.tsx
lierdakil: Um... I don't understand the question.

Pretty sure the code in `lib/main/atom/commands/refactorCode.tsx` *should* be pretty much the same as in `atom-typescript/lib/main/atom/codefix/codefixProvider.ts`.

`editor.setTextInBufferRange` in `lib/main/atom/commands/refactorCode.tsx` looks like an error to me.

lib/main/atom/commands/refactorCode.tsx
InExtremaRes: Perfect. No problemo.

lib/main/atom/commands/refactorCode.tsx
InExtremaRes: Ohh. I see, sorry, my bad. I was looking all the time at line 59. You're totally right.

lierdakil:
... I have no idea. Push it, and I'll try to have a look at some point.

InExtremaRes:
Some of the latest commits breaks the rename commando.

For instance, the next snippet

~~~ts
const things: string[] = []
const lengthOfThings = things.map(x => x.length)
~~~

If I put the cursor over `x` parameter inside callback and rename to `thing` the result is

~~~ts
const things: string[] = []
const lengthOfThings = things.map(thingthing=> x.length)
~~~

instead of the expected

~~~ts
const things: string[] = []
const lengthOfThings = things.map(thing => thing.length)
~~~

It happens in more complex code also.

InExtremaRes:
Perfect! Well I think it is working properly. What do you think?

I will try to implement this with `intentions` in the future (have little time right now) but, if I can do it, I'll make another pull request then.

Issue #1405, opened 2018-02-16
Issues with the definite assignment assertion modifier and `strictPropertyInitialization`


matt-tock:
Hi guys,

TypeScript 2.7 was released recently. They added the `strictPropertyInitialization` check, which requires the use of the `!` definite assignment assertion in some scenarios. `atom-typescript` is currently spitting out some weird errors when I'm using it. See attached images.

Any thoughts on ways I could fix this? Happy to provide more info/a simpler base case if need be. Thanks!

<img width="1107" alt="screen shot 2018-02-15 at 9 33 55 pm" src="https://user-images.githubusercontent.com/14915919/36292515-f4342d70-1297-11e8-97b5-0b320f997326.png">

<img width="1149" alt="screen shot 2018-02-15 at 9 34 23 pm" src="https://user-images.githubusercontent.com/14915919/36292524-080f87fe-1298-11e8-8c6d-ae4117484e55.png">

lierdakil:
Atom-typescript ships with typescript 2.6.2 at the moment I think, but it can use typescript 2.7 from `node_modules`. `npm install typescript` in your package directory?

matt-tock:
Ah, yea it looks like it's defaulting to 2.6.2. I'll see if I can't get it to recognize the project's typescript installation. This is slightly complicated by the fact we've actually got several projects in the same repo and the node_modules isn't at the root of the directory.

Thanks for the suggestion!

lierdakil:
Atom-typescript uses Node's [module resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together) to look for TypeScript (as implemented by [resolve package](https://www.npmjs.com/package/resolve)). So you have some options.

If all else fails, it also respects `NODE_PATH`.

lierdakil:
We finally updated bundled TypeScript to v2.7 in atom-typescript v12.4.0

PR #1419, opened 2018-04-19
Ts 2.8

lierdakil:


Neonit:
Looking forward to this one. Could need some of the new TS typing features right now. Great work!

lierdakil:
You can use whichever typescript version you want (within reason, 1.x likely won't work) with atom-typescript. See e.g. https://github.com/TypeStrong/atom-typescript/issues/1405#issuecomment-366836188

PR #1425, opened 2018-05-21
Support for insertText autocompletion suggestions

### LINKED ISSUE #1424: Support CompletionEntry insertText property
Hello, first of all great work! 

It seems you are not supporting / respecting the property `insertText` of `CompletionEntry` interface. This is returned by plugins to suggest more autocomplete entries in `LanguageService`'s method `getCompletionsAtPosition(fileName: string, position: number, options: ts_module.GetCompletionsAtPositionOptions | undefined): ts_module.CompletionInfo`

You can see the visual consequence of this issue here: https://github.com/cancerberoSgx/typescript-plugins-of-mine/tree/master/typescript-plugin-move-file . Notice that in vscode demo, the string is being replaced by a working snippet while in atom is not (and user must write it manually)

BTW I just started building plugins so we can have features like in Java / C# IDEs . The idea is also to implement this editor agnostic so they work in any TypeScript Language Service enabled editor  and to expose clear APIs to those who want to create GUIs for particular editors. Just started but already have some very useful and keep working on testing and adding missing ones. 

Thanks and keep it up!

lierdakil:

Closes #1424


cancerberoSgx:
didn't tested but code seems to be OK, thanks!

PR #1445, opened 2018-08-16
Fix rename refactor with TS 3.0

### LINKED ISSUE #1444: Atom-typescript is not renaming some symbols correctly when using TypeScript 3
For instance, in the next snippet, I put the cursor over the `a` parameter, press F2 and rename to `arg`

```ts
const foo = (a: number, b: number) => a + b;
```

After pressing Enter this is the result:

```ts
const foo = (arg: number, b: number) =arg a + b;
```

The expected result (as working in TS 2.9.2) is:

```ts
const foo = (arg: number, b: number) => arg + b;
```

I haven't tested/searching much, so I don't know if the problem is with atom-typescript or in the TS service, but same files/configuration works as expected in TS 2.9.

I am using:
 - atom-typescript: 12.6.2
 - TypeScript: 3.0.1
 - Atom: 1.29.0
 - Electron: 2.0.7

InExtremaRes:
- [x] All compiled assets are included.

The response body of the `rename` command comes with a `locs` array of `SpanGroup`, but atom-typescript assumed this array is ordered. In my tests TS 3.0 gives me the array in reverse from the one returned in TS 2.9.

Inspecting TS source code, [I found](https://github.com/Microsoft/TypeScript/blob/f2011ce7550fa852b22a8486f4db2c0679cd977a/src/server/session.ts#L1162-L1169) that the `SpanGroup` array is created from a map, so I guess the order is not guaranteed.

I fixed the issue ordering the array before applying the edits.

Nonetheless, I think it would be great if the `applyEdits` function could work for edits in any order. I believed it already works like that.

Fixes #1444


lierdakil:
Native maps actually guarantee that items are iterated over in insertion order:
> A Map object iterates its elements in insertion order — a for...of loop returns an array of [key, value] for each iteration.

Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

Besides, this is irrelevant, since actual *locations* are stored as plain arrays inside that map, so the order of *those* shouldn't be affected at all.

This looks like a bug in TypeScript to me (or at least an undocumented change)

I'll see what I can do in terms of fixing the code.

lierdakil:
Thanks for investigating, by the way!

InExtremaRes:
@lierdakil Can you reproduce the issue?

The linter warnings are present in master as well, unrelated to this changes.

lierdakil:
I figured that relying on the order TypeScript server gives the text spans to the client was not a good idea in the first place, so I just ameneded `applyEdits` to sort spans in reverse order unconditionally (it's actually pretty simple if spans are first converted to Atom `Range`s). So this should put an end to all issues of this type at the cost of a small performance penalty.

TBH, I didn't even try reproducing (no projects updated to TS 3.0 yet), but judging from your report, I would guess that TypeScript 3.0 returns rename edits in reverse order compared to TypeScript 2.9. Why that is, I have no idea an not much incentive to try to figure out. If I had to guess, consistency reasons.

PR #1491, opened 2019-02-18
Use onDidChangeText instead of onDidStopChanging for tracking buffer modification

lierdakil:

This essentially makes forcing buffer flushing unnecessary and hopefully avoids some bugs associated with asynchronicity. The flip-side is this might have a noticeable performance impact, but in my (admittedly limited) testing it did not.


lierdakil:
Merged and released as part of v13.2.1

PR #1496, opened 2019-03-28
fix: Rename input is unusably small

Xapphire13:
## Before
Screen Shot 2019-03-28 at 4 44 33 PM

## After
Screen Shot 2019-04-03 at 12 07 17 PM

### DIFF
--- a/styles/atomts-rename-view.less
+++ b/styles/atomts-rename-view.less
@@ -1,28 +1,36 @@
 .atomts-rename-view {
+  & > .block {
+    flex-direction: column;
+
+    & > * {
+      width: 100%;
+    }
+  }
+
   .options-label {
     color: @text-color-subtle;
     position: relative;
+
     .options {
       color: @text-color;
     }
   }
+
   .subtle-info-message {
     padding-left: 5px;
     color: @text-color-subtle;
   }
 
-  .find-container {
-    width: 200px;
-  }
-
   .find-container,
-  .replace-container,
-  .paths-container {
+  .paths-container,
+  .replace-container {
     display: -webkit-flex;
     -webkit-flex-direction: row;
+
     .btn-group {
       margin-left: @component-padding;
     }
+
     .editor-container {
       position: relative;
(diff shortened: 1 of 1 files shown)


lierdakil:
Uh. It shouldn't look like this. Why does it look like this?

Generally the "rename variable... etc" text should be above the input box. At least this was the case with all Atom versions up to 1.34 (as far as I can recall)

Are you using some kind of custom theme? Did this change in a recent Atom release? What's going on here?

lierdakil:
Tested with Atom 1.35.1. Everything behaves as it should:
rename variable dialog rendered as intended

It would appear something in your environment sets `display: flex; flex-direction:row` on the `block` css class. Pretty sure this is not Atom's default.

Xapphire13:
Interesting. Let me see if I can identify what is causing this, I dont use any custom themes, so it must be a bad package

Xapphire13:
I took a look and nothing is setting `block` to have `flex-direction: row`, in fact, nothing is setting the `flex-direction` style on it at all, so it's just defaulting to row.

Screen Shot 2019-04-03 at 11 56 45 AM

I could update my PR to include styling for this?

Xapphire13:
I have updated the PR with the styling required to get it looking exactly as you have it

lierdakil:
> nothing is setting block to have flex-direction: row

Why does it have `display: flex;` though? Where does that come from? Atom's default is `display: block`.

Xapphire13:
Does it not make sense to have the style explicitly set for the input, as to avoid being messed up by other extensions?

lierdakil:
The point is we don't have any control over "other extensions", and themes might include arbitrary overrides which we don't want to mess up. So with CSS the general rule I think should be "less is more". If you figure out which particular extension is messing up your stylesheets here, then we can discuss this further to resolve this particular conflict.

FWIW, `div.block` is a default Atom thing, which is supposed to have `display:block` I think, so whatever messes up your stylesheet breaks more than Atom-TypeScript.

PR #1514, opened 2019-10-15
make modifiaction in css

pasinduperera:


lierdakil:
Hi. Sorry it took a while to get to this. CI is complaining about formatting. Also, could you explain in a couple sentences what this PR is supposed to do? Because I'm not entirely sure I understand what it does exactly.

lierdakil:
Actually, I'm going to close this, since it's arguably an inconsequential change. Feel free to try to convince me otherwise.

PR #1516, opened 2019-10-19
Update “failed to install” to be more accessible

betheymc:
Added link to “failed to update” so that if the user doesn’t read the entire thing but is instead scrolling through the faqs, they know what to click to get to the correct instructions.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -109,7 +109,7 @@ instances and run the following commands:
 
 ## Failed to install
 
-Follow the same steps as specified in failed to update.
+Follow the same steps as specified in [failed to update](https://github.com/TypeStrong/atom-typescript/blob/master/docs/faq.md#failed-to-update).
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -109,7 +109,7 @@ instances and run the following commands:
 
 ## Failed to install
 
-Follow the same steps as specified in failed to update.
+Follow the same steps as specified in [failed to update](https://github.com/TypeStrong/atom-typescript/blob/master/docs/faq.md#failed-to-update).
```
lierdakil: Since it's an internal link, it'd be arguably better to omit the page URL.
```suggestion
Follow the same steps as specified in [failed to update](#failed-to-update).
```

lierdakil: Actually, will fix this on our end

lierdakil:
Moved to #1519

PR #1517, opened 2019-10-19
Grammar revision for clarity

betheymc:
When explaining how the config options specific to projects work, it got confusing when you got to the example, so I edited the text so it would be easier to understand.


lierdakil:
Moved to #1519

PR #1518, opened 2019-10-19
Made explanatory text more understandable

betheymc:
Made explanatory text under last-cursor-position more understandable and more pleasant to read.


lierdakil:
Moved to #1519

PR #1534, opened 2020-01-03
feat/extend-hyperclick

firejune:

When this option enabled, First look for `definition` then will run `find-references` when if definition is it self on `hyperclick`.


package.json
```diff
@@ -33,6 +33,13 @@
       "default": true,
       "order": 90
     },
+    "findReferencesHyperclick": {
```
lierdakil: Since hyperclick on the definition isn't all that useful, I think we may drop the config option and instead enable this unconditionally.

lib/main/atom/commands/findReferences.tsx
```diff
@@ -14,44 +15,51 @@ addCommand("atom-text-editor", "typescript:find-references", deps => ({
...
+    await handleFindReferencesResult(result, editor, deps.histGoForward)
   },
 }))
+
+export async function handleFindReferencesResult(
+  result: protocol.ReferencesResponse,
+  editor: TextEditor,
+  histGoForward: Dependencies["histGoForward"],
+): Promise<void> {
+  const refs = await Promise.all(
```
lierdakil: While you're at it, could you also get rid of this `await`? It's a holdover from way back, and it's really not needed here. Removing it will make the 'find-references' feel way more responsive. You shouldn't need to change anything else for it to work.

firejune: Ok. I did remove that `await`.

package.json
firejune: I agree that and I did drop that option. Thanks!

PR #1535, opened 2020-01-03
feat/check-all-files-on-save

firejune:

When this option enabled, It will be report all problems in current project when on file saved.


lib/main/atom/components/statusPanel/index.tsx
```diff
@@ -106,7 +106,7 @@ export class StatusPanel implements JSX.ElementClass {
   }
 
   private renderProgress(): JSX.Element | null {
-    if (this.props.progress.value < this.props.progress.max) {
+    if (this.props.buildStatus && this.props.progress.value < this.props.progress.max) {
```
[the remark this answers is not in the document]
firejune: I added that prop of `buildStatus` to fix for occurs DOM exception some times from `etch` module. But it's no longer not happened after rewrite code from your suggestion. this statement is changed. Thanks!

lierdakil:
Right. Sorry for the delay, I was travelling last week, didn't really find the time to take a look.

This looks good, but I do have a suggestion. It is generally not a good idea to rely on `atom.workspace.getActiveTextEditor()`, especially in async code, especially from a TextBuffer event (multiple editors can share the same buffer), since it can, under some hard-to-reproduce circumstances, return a wrong editor instance. That's the reason why in my "mock code" I suggested changes to `typescriptEditorPane.ts`, and not to `typescriptBuffer.ts`. So I suggest moving the code that runs `checkAllFiles` to `typescriptEditorPane.ts`, and avoiding `getActiveTextEditor` altogether.

lib/main/typescriptEditorPane.ts
```diff
@@ -47,6 +47,11 @@ export class TypescriptEditorPane {
       editor.onDidChangePath(this.checkIfTypescript),
       editor.onDidChangeGrammar(this.checkIfTypescript),
       editor.onDidDestroy(this.destroy),
+      editor.onDidSave(() => {
+        if (atom.config.get("atom-typescript.checkAllFilesOnSave")) {
```
lierdakil: Perhaps it would also be a good idea to disable `geterr` on save in `typescriptBuffer` when `checkAllFilesOnSave` is enabled? Not entirely sure myself if it is a good idea or not, but I think the previous version of the code did that.

firejune: I think doesn't need disable `geterr` when call with `{allFiles: true}` option on save. because:
- `check-all-files` are not recognized change state of text buffer when before save.
- It response faster of diagnostics messages then `check-all-files`.

PR #1541, opened 2020-01-21
Add multistep request completion event

firejune:

TS version 2.2 added a protocol of `RequestCompletedEvent`. this event as detecting multistep operation and completion commands like `geterr` and `geterrForProject`. So I add little changes of callback system in Atom-Typescript for catching multistep completion.

It's expect more accurate latency in `geterr` and  `geterrForProject` commands.


lierdakil:
Thanks for the PR! I took the liberty of pushing here directly instead of doing the review proper (also did some old code clean-up). If you could look over the changes to make sure I didn't accidentally break anything and give me a go-ahead, that would be great! A short change summary is available in the extended commit description.

firejune:
I think this is certainly a correct change in behavior 👍

Issue #1554, opened 2020-07-24
atom typescript slows down the editor in large files


aminya:
Try openning a big file and edit it and move the cursor around. Each cursor movement takes some time. Like: https://github.com/atom/atom/blob/master/src/package.js

The only cursor trackers I can think of in atom-typecsript is the SemanticsView. This is strange because my atom-ide-outline does not slow down the editor (although I have written atom-ide-outline using pure DOM, so it is expected to be faster than etch).

If you know any other, you might have to optimize the code.

lierdakil:
There's also an occurrence tracker which binds to cursor movement. That is, the thing that highlights all occurrences of the identifier under text cursor.

lierdakil:
Also, I'm not seeing the issue: https://youtu.be/TRxOH-Y28mU

aminya:
I am going to do more investigation to see which plugin is causing this!

aminya:
I found an easy way to measure things. I noticed that these functions are taking most of the time:
![image](https://user-images.githubusercontent.com/16418197/88453359-c8713780-ce2b-11ea-938b-c15ce9055737.png)


I disabled the checkAllOption in #1555, but `geterr` is slow, and it runs on every change I make in the editor.

For example, this is the result of pressing the space bar:
![image](https://user-images.githubusercontent.com/16418197/88453383-fa829980-ce2b-11ea-8fc7-2e2be7b7629e.png)

lierdakil:
Getting response for geterr request can easily take half a second, but it is an asynchronous request to a node process, so it does not block the main UI thread.

Just use the developer tools profiling tab to figure out what's going on. Measuring how long async functions take to resolve is not a meaningful metric.

aminya:
@lierdakil Here is the profile of me just typing a few things: [Profile-20200725T074758.zip](https://github.com/TypeStrong/atom-typescript/files/4975533/Profile-20200725T074758.zip)

I am seeing something very strange! I do not have atom-beautify installed, but it is shown in this profile
![image](https://user-images.githubusercontent.com/16418197/88457398-9cfe4500-ce4b-11ea-892d-555be450b286.png)


> Getting response for geterr request can easily take half a second, but it is an asynchronous request to a node process, so it does not block the main UI thread.
> 

Javascript is one-threaded in the end and async does not help much for intensive tasks. If the issue is just an external server response then async helps. I have not checked the `geterr` code, so I am not sure about it.

[...]

PR #1555, opened 2020-07-25
Clarify "Show problems for project" config

aminya:

One of the reasons that my editor slowed in #1554 was that I had enabled this option. I added a warning so people know about it. I recommend adding a tool-bar button for this command instead of tracking the atom file saving. 

This can be a huge lag for big projects. On each keyboard stroke: 800ms for Atom, and 200ms for atom-typescript itself.


lib/main/atomts.ts
```diff
@@ -28,6 +28,11 @@ export function activate(state: State) {
   pluginManager = new PluginManager(state)
 
   setImmediate(() => handlePromise(checkAndInstallDependencies()))
+  // add warning for the slow down
+  if (atom.config.get("atom-typescript.checkAllFilesOnSave")) {
```
lierdakil, line 32: This option is disabled by default. I can't get behind annoying the user with pop-ups in this case.

aminya: I wanted it to be a one-time update release note thing, but I could not find a way to do so.

Maybe something like this: https://github.com/JunoLab/atom-julia-client/blob/master/lib/package/release-note.js

lierdakil:
I've cherry-picked d5dde73 into v13.9.2, albeit I've changed the wording a little. I really don't want to annoy users with pop-ups. Those who want to, will read the changelog anyway. Those who don't, won't read the pop-up either and will get annoyed.

PR #1556, opened 2020-07-25
Fix errorPusher Performance

aminya:

takes out the loop invariant code

This snippet of the code is called tens or hundreds of times. Previously the number was multiplied by the number of diagnostics! If you can take this out from this nested loop, it would be better.
image

## After
now after this PR pushError itself takes this much time, which is ~50ms faster in each call
image

## Before
Previously errorPusher took this much:
image


lierdakil:
Something similar done in v13.9.2. Good catch.

PR #1559, opened 2020-09-26
use new atom ide types

aminya:
- [x] there is no compiled file

This uses the new atom-ide types which are provided in atom-ide-base.


aminya:
@lierdakil When I run prettier nothing changes! Is this error a false-positive?

lierdakil:
Apparently prettier broken on good ol' Trusty. I've just updated Travis config to use Bionic. It should (hopefully) work if you merge/rebase.

aminya:
Rebased it. Now it works 😄

lierdakil:
Can we have a types-only package? It feels somewhat weird to install something just for the types. Especially considering atom-ide-base npm package is 3.6M when the types we need are only 44K (i.e. about 1.2%)

Yes, I know, common wisdom is to bundle types with the package. But here it's a bit different, since the absolute majority of `atom-ide-base` consumers won't intend to actually use any _code_ in the package.

aminya:
@lierdakil 

Actually, I plan to add more functionalities to atom-ide-base. Since it is both an apm and an npm package. So in the near future, there will be some code that all the IDE packages would need to use. Many of the logics can be refactored in a commonplace.

This should not affect the end-user for now. It is just a devDependency. 🤔 Does it?

aminya:
atom-ide-base 1.4.0 will be released which is smaller. I will update this PR.

aminya:
Updated to atom-ide-base 1.4.1 which is smaller.

lierdakil:
> This should not affect the end-user for now.

It should not, unless something is very wrong, or the end-user decides they want to hack on the package. This is admittedly a minor nitpick.

> Updated to atom-ide-base 1.4.1 which is smaller.

Okay, cool, works for me.

lierdakil:
Merged manually due to conflicts.

PR #1560, opened 2020-09-26
Improve loading time by building via Parcel

aminya:

This reduces the loading time of atom-typescript (when no activation hook enabled) from 245ms to 145ms. This means the JavaScript code is loaded 100ms faster! 🚀 
ts

### made on top of #1559


lierdakil:
I've pulled the relevant parts and released as v14.0.0.

Issue #1566, opened 2020-11-17
Option "Enable Atom-Typescript for JavaScript files" cannot be disabled


alexeiz:
This option is "on" by default.  If I set it to "off" and then restart Atom, it's still in the "on" state.  The version of atom-typescript is 14.0.3.

lierdakil:
Can't reproduce.

If you're using a non-official Atom distribution, it's something that fork might be doing, take it up with the distributor.

Otherwise, post outputs of the following commands:

`apm --version`
`atom --version`
`apm list`

lierdakil:
I've searched for a bit, and apparently 
`atom-ide-javascript` package forces this option to the "on" state. If you're using it, feel free to [take it up with them](https://github.com/atom-ide-community/atom-ide-javascript/issues) but be aware that it's a "wrapper" package -- it doesn't really do much besides installing some dependencies and configuring atom-typescript, so you should be free to disable/remove it in case you don't want atom-typescript on JS.

alexeiz:
Indeed, disabling `aom-ide-javascript` resolves this issue.

PR #1575, opened 2021-02-28
SemanticView optimizations

### LINKED ISSUE #1570: Semantics view large file optimizations
I have done some optimizations to prevent freezes when the outline is going to be shown for a large file.

Since semantics-view is similar, I think the same optimizations should be used.

- skip following cursor in large files
- How long to wait for the new changes before updating the outline depends on the line count

https://github.com/atom-community/atom-ide-outline/blob/a1ec8adf5399df4fc0caecdef85ec399e4ec6343/src/main.ts#L51

P.S: I will extend the API of the outline, so you can just make the semantics-view through the outline API.

aminya:

### Description:
- Replace continuous cursor following with a command
- Large file optimization for updating outline

Fixes #1570 

In the long run, we should add the missing functionality to outline that makes you use a separate semantic view.


lib/main/atom/utils/atom.ts
```diff
@@ -97,3 +97,32 @@ async function editorTokenized(editor: Atom.TextEditor) {
     }
   })
 }
+
+// From https://github.com/atom-community/atom-ide-outline/blob/ec1a7197d63055de910562da3cc2b95fd939afc4/src/main.ts#L53
+// minimum number of line length to trigger large file optimizations
+const longLineLength =
+  (atom.config.get("linter-ui-default.longLineLength") as number | null) ?? 4000
```
lierdakil, line 104: Grabbing configuration like this from other packages can be _very_ brittle. _Especially_ on the module level. I really don't like this, pretty sure it works as intended only whenever it feels like it.

A more robust approach that doesn't require drastic architectural changes is to defer initialization until the values are actually needed.

Also could be nice to update these via `atom.config.observe`

Also not a fan that this pulls config from `linter-ui-default` which isn't necessarily installed (I still kinda support atom-ide). Perhaps we can at expose these as configuration options in atom-typescript?

lierdakil (changes requested):
First and foremost, thank you for the pull request.

I've looked it over, and I have some suggestions. I'll mark this as "request changes" for now.

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -32,7 +33,10 @@ export class NavigationTreeComponent
   constructor(public props: Props) {
     prepareNavTree(props.navTree)
     etch.initialize(this)
-    this.subscriptions.add(atom.workspace.observeActiveTextEditor(this.subscribeToEditor))
+    this.subscriptions.add(
+      atom.workspace.observeActiveTextEditor(this.subscribeToEditor),
+      atom.commands.add("atom-workspace", "typescript:reveal-cursor", this.selectAtCursorLine),
```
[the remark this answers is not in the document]
aminya: > Also, reveal cursor isn't a very descriptive name, I think.

What name do you suggest? I used the same name for Outline.

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -177,7 +186,9 @@ export class NavigationTreeComponent
     // set navTree
     await this.loadNavTree()
 
-    this.editorScrolling = editor.onDidChangeCursorPosition(this.selectAtCursorLine)
```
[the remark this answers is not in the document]
aminya: I used to debounce this in atom-ide-outline, but even with that it really slows down the editor, and so I eventually decided to get rid of it. It is better to let the user use the command whenever they need it!

https://github.com/atom-community/atom-ide-outline/commit/78a8bd17133d755781edf371e609ab965a515d25

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: "Reveal cursor" suggests something happens with the cursor, which isn't the case as far as I can tell. `reveal-in-outilne` seems more descriptive. Also adding a short description could be useful, e.g. "Reveal the symbol closest to the text cursor in outline view" or something like that.

lib/main/atom/utils/atom.ts
aminya: I used linter-ui-default config because I want to eventually make this a central config. If we start adding this to every single package, it is really inconvenient for the users. If you want I can open a PR to add this config to atom-ide-base or atom itself.

lib/main/atom/views/outline/navigationTreeComponent.tsx
```diff
@@ -138,11 +142,16 @@ export class NavigationTreeComponent
    * HELPER select the node's HTML represenation which corresponds to the
    *        current cursor position
    */
-  private selectAtCursorLine = ({newBufferPosition}: CursorPositionChangedEvent) => {
+  private selectAtCursorLine = () => {
+    const editor = atom.workspace.getActiveTextEditor()
```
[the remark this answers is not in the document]
aminya: Sure, I'll fix this

lib/main/atom/views/outline/navigationTreeComponent.tsx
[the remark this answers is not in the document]
lierdakil: Removing features is not something I'm willing to do lightly, and I doubt the performance impact is that large on smaller files (which the absolute majority of TS files are), but I probably need to run some tests myself. Which I really don't have time to do right now. Sigh. If you're adamant about this change, it'll have to wait until I manage to convince myself it's worth it.

lib/main/atom/utils/atom.ts
lierdakil: I understand the intention. However, this feels very arcane for the end-user. Here's a solution that will kind of work for now: expose these options in atom-typescript; set default to `0`, in the description explain that `0` pulls from linter-ui-default or uses `X` value if that fails.

The best option is to have this exposed as core Atom settings (well, editor atom settings more precisely), but I don't know if we want to hold this off until it gets into a public Atom release even _if_ the pull request gets merged.

In principle, I don't mind pulling from `linter-ui-default` _temporarily_, but I fear this will end up haunting the code base for the years to come, so I'm a bit reluctant.

That all said, I'm more concerned about this code being brittle.

lib/main/atom/utils/atom.ts
```diff
@@ -97,3 +97,32 @@ async function editorTokenized(editor: Atom.TextEditor) {
...
+// minimum number of lines to trigger large file optimizations
+const largeFileLineCount =
+  ((atom.config.get("linter-ui-default.largeFileLineCount") as number | null) ?? 3000) / 6
+
+export function lineCountIfLarge(editor: Atom.TextEditor) {
+  // @ts-ignore
+  if (editor.largeFileMode) {
```
lierdakil, line 111: This is always `undefined`.

lib/main/atom/utils/atom.ts
```diff
@@ -97,3 +97,32 @@ async function editorTokenized(editor: Atom.TextEditor) {
...
+  const lineCount = editor.getLineCount()
+  if (lineCount >= largeFileLineCount) {
+    // large file detection
+    return lineCount
+  } else {
+    // long line detection
+    const buffer = editor.getBuffer()
+    for (let i = 0, len = lineCount; i < len; i++) {
+      if (buffer.lineLengthForRow(i) > longLineLength) {
+        return longLineLength
```
lierdakil, line 123: Why is it returning the max length of line as a line count???

aminya:
Could you please state the missing capabilities from the outline so I can move them to the outline and just remove SemanticsView from atom-typescript? 

I recently added another set of optimizations to the outline which further increases its performance. I have also talked to some UI experts, and they have sent me a very complex level-aware lazy-rendered tree that has a virtualized scroller, and I am planning to use it inside the outline.

People are already complaining about the duplication and its performance drawbacks. I don't think this duplication is good for the ecosystem.

lierdakil:
> I don't think this duplication is good for the ecosystem.

I disagree. Competition begets innovation.

Anyway, I'm not going to agree to removing semantic view in the foreseeable future. For one, removing features isn't something to do lightly, even if those features are somewhat redundant. For two, it's not my work, and I need to have a really damn good reason to throw it out. People who prefer atom-ide's outline don't have to use semantic view, and vice versa -- the configuration literally boils down to enabling one tick and disabling another.

I've just pushed a17290dca5de7fd2cc06f95f05583b7654f60f9b to master which in my limited testing fixes the performance issues with `onDidChangeCursorPosition` event handler (it avoids costly reflows via some structure reshuffling and patches DOM directly instead of going through etch). I suggest rebasing and reverting its removal.

If you don't want to spend any more time on this pull request, let me know, I will tweak it myself when I find the time.

lib/main/atom/utils/atom.ts
lierdakil: @aminya I really need to understand this snippet. Shouldn't it return `this.largeFileLineCount` or something similar? Line length is not equivalent to line count by any stretch of the imagination.

lierdakil:
I've made some tweaks. If you could do a quick review and give me the green light that would be nice.

aminya:
> > I don't think this duplication is good for the ecosystem.
> 
> I disagree. Competition begets innovation.
> 
> Anyway, I'm not going to agree to removing semantic view in the foreseeable future. For one, removing features isn't something to do lightly, even if those features are somewhat redundant. For two, it's not my work, and I need to have a really damn good reason to throw it out. People who prefer atom-ide's outline don't have to use semantic view, and vice versa -- the configuration literally boils down to enabling one tick and disabling another.
> 

Why should we compete to achieve the very same thing while we could join forces and collaborate to make a better product that suits the needs of both projects?

I see that you have a very different perspective on the issue, which is different from how I think. 

I wanted to help this project, and that's why I spent duplicating my optimization efforts just to help SemanticsView keep up with the outline. By suggesting moving the features from this repository to outline, I wanted to help all the languages benefit from this and improve Atom as a whole not just one project.


> I've made some tweaks. If you could do a quick review and give me the green light that would be nice.

I was planning to move `lineCountIfLarge` to atom-ide-base, but now it is hardly integrated as a class method which makes this impossible. 

Given my failed efforts for helping this project, feel free to use this branch as you wish.

lierdakil:
> Why should we compete to achieve the very same thing while we could join forces and collaborate to make a better product that suits the needs of both projects?

By the same logic we should abandon Atom and switch to VSCode, because both are similar products, and we should strive to make one-size-fits-all solution instead of fragmenting the ecosystem. I hope it's pretty obvious things rarely work out this way.

> I was planning to move lineCountIfLarge to atom-ide-base, but now it is hardly integrated as a class method which makes this impossible.

Not impossible, it's actually pretty easy to factor out. As I said above, pulling from Atom config on module load is pretty bad, since config schema is still being initialized at that point, so there are no guarantees you'll get what you expect. One could, however, move all the initialization code into a factory function, and then return `lineCountIfLarge` as a `DisposableLike & { (editor: TextEditor) => number }` (i.e. a function with a `dispose` method). I can't say I'm particularly motivated to do the legwork required however, from my point of view this is used exactly once in this here code base, hence I don't see much benefit in generalizing it.

aminya:
> By the same logic we should abandon Atom and switch to VSCode, because both are similar products, and we should strive to make one-size-fits-all solution instead of fragmenting the ecosystem. I hope it's pretty obvious things rarely work out this way.

The amount of difference between two IDEs is night and day while the difference between two outline tabs is just a matter of the icon colors. 

I am sad to do this, once I implemented all the features of semantics view in the outline, I will disable SemanticsView using configs to ensure the ecosystem is not hurt by the fragmentation. I care about the end users way more than the developers.

lierdakil:
> The amount of difference between two IDEs is night and day 

Atom and VSCode are not IDEs. Both are hackable text editors. Both use the same platform, and IDE features come from plug-ins, which are reasonably flexible in both editors (more so in Atom, but that usually requires fiddling with poorly-documented internals). The primary difference is, IDE UI features and LSP compatibility have first-party support in VSCode, where in Atom it's our headache. For the end-user, the differences are in fact extremely minor, but those are enough to prefer one over the other (that said, familiarity plays a huge role as well). Of course, there are many _internal_ differences, but most of those are completely invisible to the end-user (aside from the performance issues in Atom). So I literally have no idea what you mean here.

>  once I implemented all the features of semantics view in the outline

Besides the style (which is a non-trivial difference actually), I think the only differences in terms of functionality are private/protected/public indicators and focus following the cursor? I use neither semantic view nor outline, and I wrote neither of those, so I might be missing something. Nevertheless, the former is impossible with the current API, but nothing's stopping you from changing that. And the latter you've opted to remove, so I don't know whether you're considering adding it back in.

> I will disable SemanticsView using configs

There aren't any actually. Well, except whether to show it by default. But actually, I don't mind that, provided you only disable the semantic view when the outline view is installed (and enabled). The situation where a user might want both at the same time is pretty hard to imagine. That said, https://github.com/TypeStrong/atom-typescript/issues/1566 aka https://github.com/atom-community/atom-ide-javascript/issues/14 is still an issue, so try to leave the users an out in case they want to have a weird configuration. Be warned that in case you'll decide to actively impede users from choosing to use semantic view for whatever reason, I'll be very annoyed -- "caring for the end users" by limiting their options is a farce.

